PR #3, opened 2018-09-07
Export generic_fifo_adv

bluewww:
I'm using `generic_fifo_adv`, but it doesn't seem to be listed in `src_files.yml`.


zarubaf:
Thanks. 

I strongly adivse to use the new FIFOs though ;) `generic_fifo_adv` seems like a fairly straight forward fifo.

PR #10, opened 2018-11-02
Add Popcount circuit & tests

meggiman:
Added a parametric circuit that calculates the popcount of its input using a
balanced binary adder tree.


zarubaf:
Thanks!

PR #11, opened 2018-11-06
:sparkles: Add register macros for RTL

stmach:


zarubaf:
Should we maybe prefix the macros with something unique to us? Just to avoid collisions in other projects...

PR #13, opened 2018-11-27
generic_fifo: Fix `valid_i` and `grant_i` guarantees

accuminium:
The original `generic_fifo` guaranteed that `valid_i` could be asserted even though the FIFO instance was full (i.e., `~grant_o`) and that `grant_i` could be asserted even though the FIFO instance was empty (i.e., `~valid_o`).  Before this commit, the `generic_fifo` in this repository did not fulfill these guarantees (they caused assertion violations in the internally instantiated FIFO).  This commit fixes this without changing any other properties.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add ready/valid handshake delayer
 - Add stream arbiter
 
+### Fixed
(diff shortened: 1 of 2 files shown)


accuminium:
I tested this in bigpulp, where `generic_fifo` is used at multiple places including a relatively recent PULP cluster. After this commit, that PULP cluster and SoC is compatible with the most recent `common_cells`.

zarubaf:
Seems reasonable.

accuminium:
Thanks for reviewing and approving. I will add Changelog entries and then merge the branch.

PR #14, opened 2018-12-11
Release 1.10

accuminium:
It's been a month since 1.9.0 and I need some stuff that is not yet released for the new release of the AXI repo, so I started preparing 1.10.0.

### Added
- Add `fifo_v3` with generic fill count
- Add 16 bit LFSR
- Add stream delayer
- Add stream arbiter
- Add register macros for RTL
- Add shift register

### Changed
- Make number of registers of `rstgen_bypass` a parameter.

### Fixed
- Fix `valid_i` and `grant_i` guarantees in `generic_fifo` for backward compatibility.
- Add missing entry for `popcount` to `Bender.yml`.
- Add default values for parameters to improve compatibility with Synopsys DC and Vivado.

## Points for Discussion

I see the following points to discuss, with a primary addressee (others feel free to comment ofc):

@smach: In order to reduce code duplication, I changed `stream_arbiter` to simply instantiate the `stream_arbiter_flushable` you created from the original `stream_arbiter` implementation. Do you agree?

~~@smach: I renamed and moved `header/register_defines.svh` to `include/common_cells/registers.svh` and added a `export_include_dirs` entry to `Bender.yml`. This way, any module that has `common_cells` as dependency can `include "common_cells/registers.svh"` and then use your macros. This is the convention @fabianschuiki and I envision for this feature.~~ Agreed.

~~@zarubaf: You created a `ready_valid_delay`, which would be released with 1.10.0.  I suggest to rename this module to `stream_delay` to align it with the other modules that implement the ready-valid stream interface (`stream_arbiter`, `stream_mux`, `stream_register`, etc). I did not rename it yet. What do you think?~~ Agreed, will be released as `stream_delay`.


zarubaf:
@accuminium Re-naming SGTM, better do it now than later.

accuminium:
Thanks for your feedback here and in #15, @zarubaf. I implemented both changes, so 1.10 will add a `shift_reg` and a `stream_delay` module.

fabianschuiki:
I second @stmach's concern regarding the naming of the header file. `registers.svh` seems to be more general than `flipflops.svh`.

accuminium:
Sure, sounds absolutely reasonable. I changed this.

accuminium:
@stmach: Please approve this PR if you agree with how your comment was implemented.

PR #16, opened 2018-12-13
:space_invader: Fix streaming operator in ternary for synth

stmach:
Some synthesis tools that shall not be named don't support streaming operators as ternary operands.


accuminium:
LGTM, thanks. Can you add a Changelog entry to the _Fixed_ section of 1.10.0, e.g.,
> lzc: Fix streaming operator in ternary operator for synthesis.

?

PR #21, opened 2019-02-28
Release 1.11

accuminium:
Four new modules(stream fork, stream filter, fall-through register, and ID queue) call for a new minor release! Additionally, this brings a few clean-ups and an effort to bring back continuous synthesis and simulation regression runs.

Reviewers:
- @msfschaffner mainly for the [ID queue](https://github.com/pulp-platform/common_cells/commit/78096178b5a4058925ebe191fa0176e5c5fb7f7c)
- @fabianschuiki mainly for the Bender- and CI-related improvements (https://github.com/pulp-platform/common_cells/commit/5ae503bd61ea11cf5d5e7a1f8a579b4b59a57f01, https://github.com/pulp-platform/common_cells/commit/6cf0933e6ec7209472a8c9ee05dff1aac8606f1d, https://github.com/pulp-platform/common_cells/commit/534cde5b4a02a80df62c8e6561f38b181292ed6e). Maybe you have a quick fix for the currently failing [`cdc_fifo_tb`](https://github.com/pulp-platform/common_cells/pull/21/commits/534cde5b4a02a80df62c8e6561f38b181292ed6e#diff-fbc92d81fe3afe2ef9cb1011ea14172aR24)?
- @zarubaf as overlord of this repository, for everything :wink:

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -69,6 +69,13 @@ sources:
       - test/graycode_tb.sv
       - test/stream_register_tb.sv
 
+  - target: synth_test
+    files:
+      # Level 0
+      - test/cdc_2phase_synth.sv
+      # Level 1
+      - test/synth_bench.sv
+
   # Deprecated modules
   # Level 0
   - src/deprecated/clock_divider_counter.sv
--- a/Bender.yml
+++ b/Bender.yml
@@ -46,6 +46,7 @@ sources:
   - src/cdc_fifo_gray.sv
   - src/edge_detect.sv
   - src/fifo_v2.sv
+  - src/id_queue.sv
   - src/rrarbiter.sv
   - src/rstgen.sv
   - src/stream_delay.sv
@@ -67,6 +68,7 @@ sources:
       - test/cdc_fifo_tb.sv
       - test/fifo_tb.sv
       - test/graycode_tb.sv
+      - test/id_queue_tb.sv
       - test/popcount_tb.sv
       - test/stream_register_tb.sv
 
@@ -74,6 +76,7 @@ sources:
     files:
       # Level 0
       - test/cdc_2phase_synth.sv
+      - test/id_queue_synth.sv
       # Level 1
       - test/synth_bench.sv
 
--- a/Bender.yml
+++ b/Bender.yml
@@ -48,6 +48,7 @@ sources:
   - src/rstgen.sv
   - src/stream_delay.sv
   # Level 2
+  - src/fall_through_register.sv
   - src/fifo_v1.sv
   - src/stream_arbiter_flushable.sv
   - src/stream_register.sv
--- a/Bender.yml
+++ b/Bender.yml
@@ -35,6 +35,7 @@ sources:
   - src/shift_reg.sv
   - src/spill_register.sv
   - src/stream_demux.sv
+  - src/stream_filter.sv
   - src/stream_fork.sv
   - src/stream_mux.sv
   - src/sync.sv
--- a/Bender.yml
+++ b/Bender.yml
@@ -8,6 +8,7 @@ package:
     - "Manuel Eggimann <meggimann@iis.ee.ethz.ch>"
 
 dependencies:
+  common_verification: { git: "https://github.com/pulp-platform/common_verification.git", version: 0.1.1 }
   tech_cells_generic: { git: "https://github.com/pulp-platform/tech_cells_generic.git", version: 0.1.1 }
 
 export_include_dirs:
--- a/Bender.yml
+++ b/Bender.yml
@@ -14,39 +14,44 @@ export_include_dirs:
   - include
 
 sources:
-  - src/shift_reg.sv
-  - src/fifo_v1.sv
-  - src/fifo_v2.sv
-  - src/fifo_v3.sv
-  - src/lfsr_8bit.sv
-  - src/lfsr_16bit.sv
-  - src/spill_register.sv
-  - src/stream_register.sv
-  - src/stream_mux.sv
-  - src/stream_demux.sv
-  - src/popcount.sv
+  # Source files grouped in levels. Files in level 0 have no dependencies on files in this package.
+  # Files in level 1 only depend on files in level 0, files in level 2 on files in levels 1 and 0,
+  # etc. Files within a level are ordered alphabetically.
+  # Level 0
   - src/cdc_2phase.sv
+  - src/clk_div.sv
+  - src/counter.sv
+  - src/edge_propagator_tx.sv
+  - src/fifo_v3.sv
+  - src/graycode.sv
+  - src/lfsr_16bit.sv
+  - src/lfsr_8bit.sv
+  - src/lzc.sv
+  - src/mv_filter.sv
+  - src/onehot_to_bin.sv
+  - src/popcount.sv
+  - src/rstgen_bypass.sv
+  - src/serial_deglitch.sv
+  - src/shift_reg.sv
+  - src/spill_register.sv
+  - src/stream_demux.sv
+  - src/stream_mux.sv
+  - src/sync.sv
+  - src/sync_wedge.sv
+  # Level 1
   - src/cdc_fifo_2phase.sv
   - src/cdc_fifo_gray.sv
-  - src/onehot_to_bin.sv
-  - src/rstgen.sv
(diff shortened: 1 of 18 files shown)


src/fall_through_register.sv
```diff
@@ -0,0 +1,58 @@
...
+        .clk_i          (clk_i),
+        .rst_ni         (rst_ni),
+        .flush_i        (clr_i),
+        .testmode_i     (testmode_i),
+        .full_o         (fifo_full),
+        .empty_o        (fifo_empty),
+        .alm_full_o     ( ),
+        .alm_empty_o    ( ),
+        .data_i         (data_i),
+        .push_i         (valid_i && !fifo_full),
```
zarubaf: I'd suggest you use bit-wise operations instead of logical ones e.g.: `valid_i & ~fifo_full`.

src/fall_through_register.sv
```diff
@@ -0,0 +1,58 @@
...
+        .data_o         (data_o),
+        .pop_i          (ready_i && !fifo_empty)
```
zarubaf: Same here.

src/id_queue.sv
```diff
@@ -0,0 +1,263 @@
...
+        .cnt_o      (linked_data_free_idx),
+        .empty_o    ()
+    );
+
+    // The queue is full if and only if there are no free items in the linked data structure.
+    assign full = !(|linked_data_free);
+
+    assign inp_gnt_o = !full;
+    always_comb begin
+        match_id            = 'x;
```
zarubaf: The `x` never poisons down-stream circuits, right? In general, I would like to avoid them as much as possible. Can you for example re-formulated the conditions you want to cover by using SVA?

src/stream_register.sv
```diff
@@ -46,9 +46,9 @@ module stream_register #(
         .alm_full_o     ( ),
         .alm_empty_o    ( ),
         .data_i         (data_i),
-        .push_i         (valid_i),
+        .push_i         (valid_i && !fifo_full),
```
zarubaf: Same here with bit-wise operations pls.

test/graycode_tb.sv
```diff
@@ -15,7 +15,7 @@ timeunit 1ns/1ns;
 
 module graycode_tb;
 
-    parameter int N;
+    parameter int N = 9;
```
zarubaf: Is that a parameter to the TB then I'd prefer normal port-list like parameter. If it is a `localparam` then pls re-name.

src/id_queue.sv
msfschaffner: I agree, let's avoid the `x` if possible.

src/fall_through_register.sv
msfschaffner: FYI this is suggested in our new style guide...

accuminium: The two variants have the same behavior, but as @msfschaffner said the current version is suggested by the latest style guide: "Use logical constructs for logical comparisons, bit-wise for data." and I interpreted this as a case of a logic (true/false) evaluation.

I will change it if you insist, just want to check that I am not misunderstanding something.

src/id_queue.sv
accuminium: No, these`x` should never leave this module. `match_id` can be invalid (I just added an internal `_valid` signal that was missing) and in that case it may not be used. I can bind it to `'0` instead if you prefer, but to quote the latest style guide again: "Explicitly specify don't cares (Xs) when safe to do so."

test/graycode_tb.sv
accuminium: Agreed. I did not write that TB, but I will change it.

src/fall_through_register.sv
zarubaf: I am not insisting.

src/id_queue.sv
zarubaf: If they are not leaving the module it is fine with me.

src/fall_through_register.sv
zarubaf: But I am also not agreeing. Logical comparisons for me would be usage in either `if () else ()` or a ternary expression.  You are using it as an (implicit) signal assignment, hence you are using it as data.

accuminium: Okay, I will change it and the other occurrences you mentioned.

src/id_queue.sv
```diff
@@ -0,0 +1,268 @@
...
+        match_id_valid      = 1'b0;
+        head_tail_d         = head_tail_q;
+        linked_data_d       = linked_data_q;
+        oup_gnt_o           = 1'b0;
+        oup_data_o          = data_t'('x);
```
accuminium: @zarubaf: These `x` go to an output, but they are accompanied by `oup_data_valid_o`, which is always `0` when `oup_data_o` is undefined. So `oup_data_valid_o` may only be evaluated in statements like
```sv
if (oup_data_valid_o) begin
  case (oup_data_o) ...
end
// or
if (oup_data_valid_o && oup_data_o) ...
```
Everything else is a violation of the interface specification and propagating `x` makes verification and debugging easier in this case IMO. Do you agree?

accuminium:
Another change that should make it into 1.11: @SamuelRiedel noticed and fixed a bug in `fifo_v3` where data did not propagate through an empty fall-through FIFO (7f387051a80f91afa65a8f4151da481bd2ec8bfb). To verify the change, I have updated `fifo_tb` to work with the requirements on `push` and `pop` that have been in use for some time now (a71d87a4046c194920225c59411ada01d4077735), extended the test cases (ebaca4bba0159f71a423c9af1b4e7b68ea0c2d58), and added an assertion for the fall-through property as regression test (a340f6f5fdcbc3ed6bcae190e65c31340d57d5cc).

@zarubaf: Do you agree with this change? Could you please also give a brief comment on the open discussion above?

@fabianschuiki: Could you please check the changes I mentioned for you in the description?

Thanks!

src/id_queue.sv
zarubaf: So to be honest I am missing the expertise on this. My experience so far tells me:

- `x` usually brick assertions and other run-time checks in assertions. But I also see that the downstream circuit would violate the protocol if it would unconditionally check the output. 
- I highly doubt that you will see any advantage in synthesis. Of course, the only way to find out is to actually synthesize the circuit with both conditions.

I personally would not use `x` in synthesizable circuits, I've been burned a couple of times. But I will go with whatever decision you are making. No hard feelings ;-)

zarubaf:
Good catch. Definitely, an undesired condition. Thanks @SamuelRiedel

src/id_queue.sv
accuminium: Thanks for giving your view on this. Then I would say we give it a try and stick to the style guide when it is safe to do so (in this case given by the accompanying `_valid` signal). There might be no advantages for synthesis, but it has advantages in verification and debugging.

zarubaf:
Ping @msfschaffner @fabianschuiki

Bender.yml
```diff
@@ -8,63 +8,89 @@ package:
     - "Manuel Eggimann <meggimann@iis.ee.ethz.ch>"
 
 dependencies:
+  common_verification: { git: "https://github.com/pulp-platform/common_verification.git", version: 0.1.1 }
```
fabianschuiki: I like the separate collection of verification tools!

src/stream_register.sv
```diff
@@ -46,9 +46,9 @@ module stream_register #(
         .alm_full_o     ( ),
         .alm_empty_o    ( ),
         .data_i         (data_i),
-        .push_i         (valid_i),
+        .push_i         (valid_i & ~fifo_full),
```
fabianschuiki: This should get rid of some annoying assertions triggering :smile:

accuminium:
Thanks for your reviews!

> I know this is nitpicking, but our style guide would recommend to use `~` here as you operate on data.

Okay, fixed.

accuminium:
I have rebased this branch on `master` to remove conflicts in `CHANGELOG.md`. I have removed the `## v1.11.0` line (previously in 87029835e4be68ebab18ce23ffa0484ca80d2f26) because the date was updated and I think the owner (@zarubaf) should add that line before creating the release tag.

All simulation and synthesis tests pass, so ready to merge from my view.

PR #22, opened 2019-03-01
sram: Add FPGA cpabilities and random initialization

zarubaf:
I was wondering if I can sneak in another change into the `1.11` CL? This one adds improved SRAM capabilities. In particular it adds:

1. sim initialization: either don't initialize, zero, random values or the usual `0xdeadbeef`
2. In an FPGA setting, they will properly be inferred as BRAMs. The drawback that this adds a dependency to the `fpga-support` module. But imho it is very convenient to have inferrable SRAMs.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -10,6 +10,7 @@ package:
 dependencies:
   common_verification: { git: "https://github.com/pulp-platform/common_verification.git", version: 0.1.1 }
   tech_cells_generic: { git: "https://github.com/pulp-platform/tech_cells_generic.git", version: 0.1.1 }
+  fpga-support:       { git: "https://github.com/pulp-platform/fpga-support.git",       version: 0.3.2 }
 
 export_include_dirs:
   - include
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,8 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 - Add fall-through register
 - Add stream filter
 - Add ID queue
+- Add randomization capabilities to SRAM `sram.sv`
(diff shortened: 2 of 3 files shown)


accuminium:
I agree with the first point, but depending on a more specific repository (`fpga-support`) from a generic one (`common_cells`) looks like an anti-pattern to me.

I see the convenience of inferrable SRAMs but that is more of a technology-agnostic "macro cell" than a behavioral "common cell" to me. I see two options that involve moving the inferrable SRAM code from `fpga-support` to a more generic repo:
1. We remove `sram` from `common_cells` and create a new `sram` module in `tech_cells_generic`. That module internally instantiates a 32-bit or a 64-bit RAM through the code that currently is in `fpga-support`. To avoid code duplication, we remove that code from `fpga-support`, let it depend on `tech_cells_generic`, and simply instantiate a properly parametrized `sram` there.
2. We keep `sram` in `common_cells` but move the code from `fpga-support`, let it depend on `tech_cells_generic`, and instantiate a properly parametrized `sram`.

In my eyes, (1) puts more things where they belong, but (2) is backwards-compatible and less work. What do you think?

zarubaf:
The first option sounds good to me. I think right now Ariane should be the only repository where we use such SRAM cut and we can use the time to break backwards compatibility. I will do this beginning next week if you can wait that long before merging `1.11`.

accuminium:
Since this is a breaking change, it would go into a new major release. I think releasing `2.0` shortly after `1.11` is not a problem, though. We could use the opportunity and remove all deprecated modules.

zarubaf:
Closing this for now as for proposed solution in `tech_cells_generic`.

PR #32, opened 2019-04-20
lzc: Fix misbehaving Verilator alternative

zarubaf:
The alternative Verilator implementation was diverging from the actual
LZC implementation. This has lead to simulation mismatches.


zarubaf:
@msfschaffner Can you make sure my assumption on the interface is correct? This probably also explains the Verilator FPU mismatches we've seen but never really investigated.

zarubaf:
Yes, I'll check that.

msfschaffner:
I checked and it looks like this alternative Verilator implementation is not required anymore. The simulation speed does not suffer with this implementation (it looks like the speed loss can be attributed to the streaming operator in this case).

PR #37, opened 2019-05-09
stream_arbiter: Use `rr_arb_tree` instead of deprecating

accuminium:
I advocate to keep `stream_arbiter` as active module in the `stream_` series of ready/valid modules even with the new `rr_arb_tree`. I think it is very convenient to have a single module to instantiate to arbitrate any ready/valid stream without having to set the right parameters to `rr_arb_tree` every time. As `stream_arbiter` simply instantiates `rr_arb_tree` with the proper parameters, the maintenance cost of this module is near zero.


zarubaf:
@msfschaffner What was the rationale of removing?

msfschaffner:
The rationale was to deprecate everything that `just` wraps a newer and better implementation (like in case of the fifo variants). But I see that the wrapping module can make sense if you have a series of corresponding blocks with similar naming and protocol conventions - so I am ok with keeping it alive...

zarubaf:
Okay. Then we just need to add the priority arbitration to the module again and can merge.

msfschaffner:
Priority arbitration can be done using `rr_arb_tree.sv` by setting the `rr_i` flag constantly to `'0` (needs to be double checked).

msfschaffner:
Ah and `ExtPrio` needs to be set to `1'b1` in that case.

accuminium:
Okay, I have added the `ARBITER` parameter again, instantiating `rr_arb_tree` as you described. Additionally, I added a simple synthesis bench for regression testing, which passes.

Ready to merge from my side.

PR #42, opened 2019-05-28
Add cf_math_pkg: Constant Function Implementations of Mathematical Functions for HDL Elaboration

accuminium:
This package contains a collection of mathematical functions that are commonly used when defining the value of constants in HDL code.  These functions are implemented as Verilog constants functions.  Introduced in Verilog 2001 (IEEE Std 1364-2001), a constant function (§ 10.3.5) is a function whose value can be evaluated at compile time or during elaboration.  A constant function must be called with arguments that are constants.

### DIFF
--- a/src/cf_math_pkg.sv
+++ b/src/cf_math_pkg.sv
@@ -1,0 +1,64 @@
+// Copyright 2016 ETH Zurich and University of Bologna.
+// Copyright and related rights are licensed under the Solderpad Hardware
+// License, Version 0.51 (the "License"); you may not use this file except in
+// compliance with the License.  You may obtain a copy of the License at
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+// cf_math_pkg: Constant Function Implementations of Mathematical Functions for HDL Elaboration
+//
+// This package contains a collection of mathematical functions that are commonly used when defining
+// the value of constants in HDL code.  These functions are implemented as Verilog constants
+// functions.  Introduced in Verilog 2001 (IEEE Std 1364-2001), a constant function (§ 10.3.5) is a
+// function whose value can be evaluated at compile time or during elaboration.  A constant function
+// must be called with arguments that are constants.
+
+package automatic cf_math_pkg;
+
+    // Ceiled Division of Two Natural Numbers
+    //
+    // Returns the quotient of two natural numbers, rounded towards plus infinity.
+    function integer ceil_div(input longint dividend, input longint divisor);
+        automatic longint remainder;
+
+        // pragma translate_off
+        if (dividend < 0)
+            $fatal(1, "Dividend %0d is not a natural number!", dividend);
+
+        if (divisor < 0)
+            $fatal(1, "Divisor %0d is not a natural number!", divisor);
+
+        if (divisor == 0) begin
+            $fatal(1, "Division by zero!");
+        end
+        // pragma translate_on
+
+        remainder = dividend;
+        for (ceil_div = 0; remainder > 0; ceil_div = ceil_div+1)
+            remainder = remainder - divisor;
+    endfunction
+
+    // Ceiled Binary Logarithm of a Natural Number
+    //
+    // Returns the binary logarithm (i.e., the logarithm to the base 2) of a natural number
+    // (including 0), rounded towards plus infinity.
+    //
+    // Use this as drop-in replacement for the `$clog2` system function where the latter is not
+    // supported by your toolchain.
+    function integer log2 (input longint val);
+        automatic longint tmp;
+
+        // pragma translate_off
+        if (val < 0)
+            $fatal(1, "Argument %0d is not a natural number!", val);
+        // pragma translate_on
+
+        tmp = val - 1;
+        for (log2 = 0; tmp > 0; log2 = log2+1)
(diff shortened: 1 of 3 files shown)


src/cf_math_pkg.sv
```diff
@@ -0,0 +1,64 @@
...
+// functions.  Introduced in Verilog 2001 (IEEE Std 1364-2001), a constant function (§ 10.3.5) is a
+// function whose value can be evaluated at compile time or during elaboration.  A constant function
+// must be called with arguments that are constants.
+
+package automatic cf_math_pkg;
+
+    // Ceiled Division of Two Natural Numbers
+    //
+    // Returns the quotient of two natural numbers, rounded towards plus infinity.
+    function integer ceil_div(input longint dividend, input longint divisor);
```
zarubaf: Style: Space after `div`

src/cf_math_pkg.sv
```diff
@@ -0,0 +1,64 @@
...
+        if (divisor < 0)
+            $fatal(1, "Divisor %0d is not a natural number!", divisor);
+
+        if (divisor == 0) begin
+            $fatal(1, "Division by zero!");
+        end
+        // pragma translate_on
+
+        remainder = dividend;
+        for (ceil_div = 0; remainder > 0; ceil_div = ceil_div+1)
```
zarubaf: This needs `begin`, `end`

src/cf_math_pkg.sv
```diff
@@ -0,0 +1,64 @@
...
+
+// cf_math_pkg: Constant Function Implementations of Mathematical Functions for HDL Elaboration
+//
+// This package contains a collection of mathematical functions that are commonly used when defining
+// the value of constants in HDL code.  These functions are implemented as Verilog constants
+// functions.  Introduced in Verilog 2001 (IEEE Std 1364-2001), a constant function (§ 10.3.5) is a
+// function whose value can be evaluated at compile time or during elaboration.  A constant function
+// must be called with arguments that are constants.
+
+package automatic cf_math_pkg;
```
zarubaf: What is an `automatic` package?

src/cf_math_pkg.sv
```diff
@@ -0,0 +1,64 @@
...
+    function integer log2 (input longint val);
+        automatic longint tmp;
+
+        // pragma translate_off
+        if (val < 0)
+            $fatal(1, "Argument %0d is not a natural number!", val);
+        // pragma translate_on
+
+        tmp = val - 1;
+        for (log2 = 0; tmp > 0; log2 = log2+1)
```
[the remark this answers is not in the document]
accuminium: `begin .. end` and `log2++`: Fixed in latest commit.
Definition of `log2`: This is the return value of the function (same name as function).
`$clog2`: We added this in 2015 due to deficiencies of Vivado if I remember correctly. But recent versions of Vivado seem to implement in `$clog2` for the few use cases I briefly tested today. So if you want, I can remove the `log2` function from this package.

src/cf_math_pkg.sv
accuminium: When applied to a package, the `automatic` lifetime qualifier causes all arguments and variables of
functions and tasks to be `automatic` unless they are explicitly declared `static`.

accuminium:
Yes, this is synthesizable in DC (I tried 2018.06).

Regarding the necessity: `log2` as in the discussion above; `ceil_div` to compute how many multiples of a number you need to fit another number; e.g., given a parametric number of inputs, how many N-input blocks do I need to instantiate to process the inputs in parallel?

src/cf_math_pkg.sv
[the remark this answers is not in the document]
zarubaf: If that doesn't create problems for you I'd indeed prefer to remove that.

src/cf_math_pkg.sv
```diff
@@ -0,0 +1,69 @@
...
+            $fatal(1, "Divisor %0d is not a natural number!", divisor);
+        end
+
+        if (divisor == 0) begin
+            $fatal(1, "Division by zero!");
+        end
+        // pragma translate_on
+
+        remainder = dividend;
+        for (ceil_div = 0; remainder > 0; ceil_div++) begin
```
zarubaf: `(dividend + divisor - 1) / divisor` should do the trick here. I am not a big fan of this iterative description.

src/cf_math_pkg.sv
```diff
@@ -0,0 +1,69 @@
...
+
+package automatic cf_math_pkg;
+
+    // Ceiled Division of Two Natural Numbers
+    //
+    // Returns the quotient of two natural numbers, rounded towards plus infinity.
+    function integer ceil_div (input longint dividend, input longint divisor);
+        automatic longint remainder;
+
+        // pragma translate_off
```
zarubaf: please add an `ifdef VERILATOR` around that as well :)

src/cf_math_pkg.sv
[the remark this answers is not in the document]
accuminium: Okay, done. For the few cases I really need it (old FPGA designs), I can create local helper functions.

src/cf_math_pkg.sv
accuminium: Yes, but that has the drawback that it overflows if `dividend + divisor > 2**63`. This might sound unrealistic at first, but consider the case where you want to chunk a 64-bit address space into large segments.

I do not praise the (lacking) beauty of the iterative approach, just wanted to mention the drawback of the direct-calculation approach. I can change it if you prefer.

src/cf_math_pkg.sv
accuminium: You mean `ifndef`, right? Done.

zarubaf: Yes, sry

zarubaf:
LGTM

PR #46, opened 2019-06-01
clk_div: Fix that counter was never reset

zarubaf:


andreaskurth (approved):
The bugfix looks good to me, but there are two other things in this module that do not make sense to me:

1. When dividing by an integer factor `RATIO`, why does the counter need `RATIO` bits instead of `$clog2(RATIO)` bits?
2. Is it on purpose that the output clock has a duty cycle < 50% (in fact, just one cycle of the input clock)?

Xavier2979:
> The bugfix looks good to me, but there are two other things in this module that do not make sense to me:
> 
> 1. When dividing by an integer factor `RATIO`, why does the counter need `RATIO` bits instead of `$clog2(RATIO)` bits?
> 2. Is it on purpose that the output clock has a duty cycle < 50% (in fact, just one cycle of the input clock)?

niwis:
This is a pretty old PR, but it looks to me like this issue still persists. I also agree with the observations of @accuminium regarding the counter width and duty cycle. Do you think we should proceed on this? In that case I could propose a fix on top of this PR.

accuminium:
> Do you think we should proceed on this? In that case I could propose a fix on top of this PR.

That would be great! :+1:

niwis:
Superseded by https://github.com/pulp-platform/common_cells/pull/132

PR #47, opened 2019-06-03
rr_arb_tree assertions

suehtamacv:
I worked on the assertions of the `rr_arb_tree` module, after encountering a few violations (especially with the `LOCK_IN` feature turned on). After checking the module with a formal verification tool, more problems appeared, as follows:

- Given the `always_ff` blocks, I believe the correct reset signal is `(!rst_ni) || (flush_i)`, otherwise the properties would leave the value of `flush_i` floating. A default reset signal avoids repetition of this expression.
- Property `lock_req` seems more like an assumption than an assertion, as it controls the behavior of signals outside the module (through `req_tmp`).
- Possible typo in `req1`.

### DIFF
--- a/src/rr_arb_tree.sv
+++ b/src/rr_arb_tree.sv
@@ -55,6 +55,12 @@ module rr_arb_tree #(
   output DataType                          data_o,
   output logic [$clog2(NumIn)-1:0]         idx_o
 );
+
+  // pragma translate_off
+  // Default SVA reset
+  default disable iff (!rst_ni || flush_i);
+  // pragma translate_on
+
   // just pass through in this corner case
   if (NumIn == unsigned'(1)) begin
     assign req_o    = req_i[0];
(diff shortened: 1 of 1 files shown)


src/rr_arb_tree.sv
```diff
@@ -215,27 +221,27 @@ module rr_arb_tree #(
...
         else $fatal (1, "Idx_o / gnt_o do not match.");
 
     req0 : assert property(
-      @(posedge clk_i) disable iff (!rst_ni) |req_i |-> req_o)
+      @(posedge clk_i) |req_i |-> req_o)
         else $fatal (1, "Req in implies req out.");
 
     req1 : assert property(
-      @(posedge clk_i) disable iff (!rst_ni) |req_o |-> req_i)
+      @(posedge clk_i) req_o |-> |req_i)
```
msfschaffner: both solutions work fine as we can test implicitly for a nonzero value

PR #49, opened 2019-07-02
fix altera quartus compile error

zhangy-thu:
Change localparam to parameter to fix quartus compile error. This error also has been discussed in https://stackoverflow.com/questions/29717467/modelsim-wrong-scope-for-localparam

### DIFF
--- a/src/stream_demux.sv
+++ b/src/stream_demux.sv
@@ -17,7 +17,7 @@
 module stream_demux #(
   parameter integer N_OUP = 1,
   /// Dependent parameters, DO NOT OVERRIDE!
-  localparam integer LOG_N_OUP = $clog2(N_OUP)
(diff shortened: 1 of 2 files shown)


zarubaf:
Is this the only problem with quartus you are facing, if so I am willing to change this. If there are other issues (I remember Quartus to be very/unreasonably picky on SV) I'd prefer to keep this as is. Your input is much appreciated.

zhangy-thu:
For the common_cells, this is the only issue for Quartus (Yes, Quartus is very unfriendly for SV, while the latest release have some improvements). I have completed fpga verification for Ariane and PULPissimo based on Altera A10 fpga.

zarubaf:
Thanks! Then I am going ahead merging this!

PR #51, opened 2019-07-30
[onehot-to-bin] Fix immediate assertion

suehtamacv:
The immediate assertion in `onehot_to_bin.sv` was inside an initial block, and was only checked at the beginning of the simulation.

Because the module has no clock, I added the "final" attribute to postpone the evaluation of the assertion until the signals settle. I checked it with ModelSim 10.6, but previous versions of the simulator might have issues with final deferred assertions.


zarubaf:
Thanks! Lgtm

PR #56, opened 2019-10-17
Quartus support

MarekPikula:
In reference to: pulp-platform/pulpissimo#78

Fixed issues:

- [type parameters](https://marekpikula.github.io/quartus-sv-gotchas/Intel%20Quartus%20SystemVerilog%20gotchas.html#6203-type-parameters)
- [conditional generate](https://marekpikula.github.io/quartus-sv-gotchas/Intel%20Quartus%20SystemVerilog%20gotchas.html#275-conditional-generate-constructs)
- [loop generate](https://marekpikula.github.io/quartus-sv-gotchas/Intel%20Quartus%20SystemVerilog%20gotchas.html#274-loop-generate-constructs-3)
- changed Vivado's `dont_touch` synthesis flag to Quartus' `preserve keep`.

### DIFF
--- a/src/cdc_2phase.sv
+++ b/src/cdc_2phase.sv
@@ -17,28 +17,30 @@
 /// the paths async_req, async_ack, async_data.
 /* verilator lint_off DECLFILENAME */
 module cdc_2phase #(
-  parameter type T = logic
+  parameter int unsigned T_w = 1
 )(
   input  logic src_rst_ni,
   input  logic src_clk_i,
-  input  T     src_data_i,
+  input  logic [T_w-1:0] src_data_i,
   input  logic src_valid_i,
   output logic src_ready_o,
 
   input  logic dst_rst_ni,
   input  logic dst_clk_i,
-  output T     dst_data_o,
+  output logic [T_w-1:0] dst_data_o,
   output logic dst_valid_o,
   input  logic dst_ready_i
 );
 
+  typedef logic [T_w-1:0] T;
+
   // Asynchronous handshake signals.
-  (* dont_touch = "true" *) logic async_req;
-  (* dont_touch = "true" *) logic async_ack;
-  (* dont_touch = "true" *) T async_data;
+  logic async_req /* synthesis keep preserve */;
+  logic async_ack /* synthesis keep preserve */;
+  T async_data /* synthesis keep preserve */;
 
   // The sender in the source domain.
-  cdc_2phase_src #(.T(T)) i_src (
+  cdc_2phase_src #(.T_w(T_w)) i_src (
     .rst_ni       ( src_rst_ni  ),
     .clk_i        ( src_clk_i   ),
     .data_i       ( src_data_i  ),
@@ -50,7 +52,7 @@ module cdc_2phase #(
   );
 
   // The receiver in the destination domain.
-  cdc_2phase_dst #(.T(T)) i_dst (
+  cdc_2phase_dst #(.T_w(T_w)) i_dst (
     .rst_ni       ( dst_rst_ni  ),
     .clk_i        ( dst_clk_i   ),
     .data_o       ( dst_data_o  ),
@@ -66,22 +68,22 @@ endmodule
 
 /// Half of the two-phase clock domain crossing located in the source domain.
 module cdc_2phase_src #(
-  parameter type T = logic
+  parameter int unsigned T_w = 1
 )(
   input  logic rst_ni,
   input  logic clk_i,
-  input  T     data_i,
+  input  logic[T_w-1:0] data_i,
   input  logic valid_i,
   output logic ready_o,
   output logic async_req_o,
   input  logic async_ack_i,
-  output T     async_data_o
+  output logic[T_w-1:0] async_data_o
 );
 
-  (* dont_touch = "true" *)
-  logic req_src_q, ack_src_q, ack_q;
-  (* dont_touch = "true" *)
-  T data_src_q;
+  typedef logic[T_w-1:0] T;
+
+  logic req_src_q /* synthesis keep preserve */, ack_src_q /* synthesis keep preserve */, ack_q /* synthesis keep preserve */;
+  T data_src_q /* synthesis keep preserve */;
 
   // The req_src and data_src registers change when a new data item is accepted.
   always_ff @(posedge clk_i or negedge rst_ni) begin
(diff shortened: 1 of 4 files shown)


zarubaf:
I see, sorry I didn't pay attention to that. What is the final goal for this branch? Who is going to maintain this?

MarekPikula:
As mentioned the final goal is to provide support for PULPissimo for Quartus, so that other people wanting to run it on this platform don't need to go through what I went through.

I can see no problem maintaining it. Any time there will be a version bump planned for PULPissimo (as there is now) I can test it and verify everything is synthesizing correctly.

zarubaf:
I understood what you want to achieve is. My question was on what your final goal for the Quartus branch is. For me, it is definitely a non-goal to maintain an out-of-tree branch. I would suggest that instead of hacking away we find another alternative to this. If you really want to keep your own version I'd suggest forking this repository.

MarekPikula:
> I understood what you want to achieve. My question was on what your final goal for the Quartus branch is.

At the moment I'm an intern in Trinamic Motion Control. My goal is to compare alternative RISC-V cores and platforms to the one that is currently used. The problem is that the current core was prototyped on Intel tools, so the only way to properly compare it was to make PULPissimo work with Intel tools. I don't know where Trinamic will go from there, so I don't know if it will find its way to silicon or not.

> For me, it is definitely a non-goal to maintain an out-of-tree branch.

Actually it's not out-of-tree branch, but one coupled with tag that PULPissimo will be based on after pulp-platform/pulp_soc#12 is merged. 

> I would suggest that instead of hacking away we find another alternative to this.

In my view the current solution to the problem is indeed one big hack, but I see no other way of introducing support for modules in `common_cells`, which are required by other modules within PULPissimo. Quartus simply doesn't support type parameters and it will not change any time soon if at all.

> If you really want to keep your own version I'd suggest forking this repository.

Sure I can and will keep my own version in fork, but in my opinion it will introduce a huge mess in the final configuration, where some IPs will be based on official PULP repos and some based on my forks.

I think that the easiest way of enabling use of Quartus for anyone wanting to try it is:
- integrate as many non-breaking changes to relevant IP repos (you can find a list and relevant PRs in pulp-platform/pulpissimo#78), so they are in main repos and there is no action required from me to update them to versions used in PULPissimo,
- create separate branches for breaking changes as it is done in this PR, which I will maintain (so every time there is a version bump for PULPissimo I get informed on the versions, which will be used, I'm checking the compatibility and pushing to quartus branch) – the only repos requiring breaking changes are `common_cells`, `riscv-dbg` and `pulpissimo` itself,
- add a section in PULPissimo readme abut possibility to use it with Quartus and guide how to do it with additional information/guide for anyone wanting to port some additional IPs to ease their pain.

accuminium:
Independent of whether such tool-specific customizations are maintained on a branch of the upstream repository or in a fork, I think a central drawback of creating variants of central modules like the ones in `common_cells` is that this entails a fragmentation of the modules that depend on them.  Keeping dependencies aligned is work enough already, but propagating updates through variants of dependencies is a lot of additional work.

An alternative that prevents this could be to use tool-specific defines, e.g.,
```sv
module cdc_2phase #(
`ifdef TARGET_QUARTUS
  parameter int unsigned T_w = 1
`else
  parameter type T = logic
`endif
)(
  input  logic src_rst_ni,
  input  logic src_clk_i,
`ifdef TARGET_QUARTUS
  input  logic [T_w-1:0] src_data_i,
`else
  input  T     src_data_i,
`endif
// etc
```

This way, the customization could be maintained inside the main upstream code. The drawback is increased code clutter that distracts from the actual logic.

While I would prefer this alternative, I cannot answer the question of whether it's worth it. Type parameters are not part of the disputed or ambiguous aspects of the SystemVerilog standard, the vast majority of EDA tools support them, and their use is accepted or even encouraged by the HDL community.  So maybe Quartus developers should just do their homework here and follow the standard like the majority.

MarekPikula:
Although Quartus Standard/Lite is still supported it doesn't come with feature updates. Intel focuses now on Quartus Pro, which supports only the more powerful (and expensive) Intel devices and lacks support for still commonly used families (like Cyclone V, which I've been using) and is relatively expensive. Quartus Pro indeed supports full SV, I presume with all these features, which I'm addressing in my PRs, but it would be of no use in my particular case.

It is clearly stated in feature comparison section in [download](https://www.intel.com/content/www/us/en/software/programmable/quartus-prime/download.html) page that SV support for both Standard and Lite versions is limited. From what I have seen over recent updates, there are no feature updates, only some library updates and small bug fixes.

It is not in my scope to nag Intel to do something about it, as I presume it would be fruitless, unless I had a lot of money to pay them to implement this. I guess customers who have a lot of money don't even bother with these small devices and thus Quartus Standard/Lite, so I would assume that scope of support for SV will not change any time soon, if at all.

In my opinion solution that you provided (@accuminium) is indeed cluttered and should be avoided if possible, but if you want this in your main repo I can prepare appropriate patches and update this PR.

MarekPikula:
There is similar discussion in lowRISC/ibex#117.

MarekPikula:
Also it's worth mentioning, that `riscv-dbg` has even more issues, most notable use of `unique case inside` statements. I have patches, but they are significant enough that making them into the main branch is in my opinion not feasible in any way and would make the code very cluttered if `ifdef` statements would be used.

zarubaf:
Okay, thanks for all the effort you put into trying to make this work. That must be one frustrating thing. I think I've found a reasonable (automated) solution to this problem:

1.  It seems that all the files you are concerned about can be translated to Verilog using `sv2v` converter (https://github.com/zachjs/sv2v). Mentioned in the ibex thread.
2.  We can add a `WIDTH` parameter in the `cdc_2phase` version which set the default type to `logic[WIDTH-1:0]`. Similar to what we've done in `fifo_v3`.
3. We add a Github Action which automatically translates the source files to verilog and maintaine a `verilog` branch.

This relieves us and you of maintaining a separate branch which I really want to avoid. This is all based on the assumption that `sv2v` does a correct translation which I can't guarantee.

MarekPikula:
Unfortunately sv2v doesn't provide converters for all unsupported constructs, for example `unique case inside`, which is a problem for `risc-dbg`. It would be indeed ideal if sv2v was working for as many components as possible. I'll check if at least `common_cells` will work with it.

PR #57, opened 2019-11-25
lzc: handle degenerate case when WIDTH=1

suehtamacv:
The LZC uses `$clog2(WIDTH)` throughout the implementation, which raises some problems when `WIDTH = 1`. This version generates a degenerated version of the LZC for this special case.

### DIFF
--- a/src/lzc.sv
+++ b/src/lzc.sv
@@ -26,68 +26,81 @@
 module lzc #(
   /// The width of the input vector.
   parameter int unsigned WIDTH = 2,
-  parameter bit          MODE  = 1'b0 // 0 -> trailing zero, 1 -> leading zero
+  parameter bit          MODE  = 1'b0, // 0 -> trailing zero, 1 -> leading zero
+  // Dependent parameters. Do not change!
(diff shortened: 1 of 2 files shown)


src/lzc.sv
```diff
@@ -26,68 +26,81 @@
...
-  input  logic [WIDTH-1:0]         in_i,
-  output logic [$clog2(WIDTH)-1:0] cnt_o,
-  output logic                     empty_o // asserted if all bits in in_i are zero
+  input  logic [WIDTH-1:0]     in_i,
+  output logic [CNT_WIDTH-1:0] cnt_o,
+  output logic                 empty_o // asserted if all bits in in_i are zero
 );
 
-  localparam int unsigned NUM_LEVELS = $clog2(WIDTH);
+  generate
```
accuminium: Please remove the `generate` .. `endgenerate` region (see https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md#generate-constructs).

accuminium:
@zaruba I think this is ready to be merged.

PR #58, opened 2019-12-02
Address decoder

WRoenninger:
Addition of an address map to index decoder module

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ Please note that cells with status *deprecated* are not to be used for new desig
 
 | Name                         | Description                                                                    | Status         | Superseded By |
 | :--------------------------- | :----------------------------------------------------------------------------- | :------------- | :------------ |
+| `addr_decode   `             | Address map decoder                                                    | active         |               |
 | `binary_to_gray`             | Binary to gray code converter                                                  | active         |               |
 | `find_first_one`             | Leading-one finder / leading-zero counter                                      | *deprecated*   | `lzc`         |
 | `gray_to_binary`             | Gray code to binary converter                                                  | active         |               |
--- a/README.md
+++ b/README.md
@@ -53,7 +53,7 @@ Please note that cells with status *deprecated* are not to be used for new desig
 
 | Name                         | Description                                                                    | Status         | Superseded By |
 | :--------------------------- | :----------------------------------------------------------------------------- | :------------- | :------------ |
-| `addr_decode   `             | Address map decoder                                                    | active         |               |
+| `addr_decode   `             | Address map decoder                                                            | active         |               |
 | `binary_to_gray`             | Binary to gray code converter                                                  | active         |               |
 | `find_first_one`             | Leading-one finder / leading-zero counter                                      | *deprecated*   | `lzc`         |
 | `gray_to_binary`             | Gray code to binary converter                                                  | active         |               |
--- a/src/addr_decode.sv
+++ b/src/addr_decode.sv
@@ -29,7 +29,7 @@
 
 // Assertions: The module checks every time there is a change in the address mapping
 // if the resulting map is valid. It fatals if `start_addr` is higher than `end_addr`
-// or if a mapping targets a port index that is outside the number of masters.
+// or if a mapping targets a port index that is outside the number of allowed indices.
 // It issues warnings if the address regions of any two mappings overlap.
 
 module addr_decode #(
@@ -45,7 +45,7 @@ module addr_decode #(
   output idx_t                idx_o,            // decoded index
   output logic                dec_valid_o,      // decode is valid
   output logic                dec_error_o,      // decode is not valid
-  // Default slave enable
+  // Default index mapping enable
   input  logic                en_default_idx_i, // enable default port mapping
   input  idx_t                default_idx_i     // default port index
 );
@@ -90,7 +90,7 @@ module addr_decode #(
   // effectively ordered. Only one of the rules with the same function is
   // active at a time for a given pair.
   // check_start:        Enforces a smaller start than end address.
-  // check_idx:          Enforces a valid master port index in the rule.
+  // check_idx:          Enforces a valid index in the rule.
   // check_overlap:      Warns if there are overlapping address regions.
   for (genvar i = 0; i < NoRules; i++) begin : gen_assert_0
     check_start : assume final (addr_map_i[i].start_addr < addr_map_i[i].end_addr) else
@@ -101,7 +101,7 @@ module addr_decode #(
           i ,addr_map_i[i].idx, addr_map_i[i].start_addr, addr_map_i[i].end_addr));
     // check the SLV ids
     check_idx : assume final (addr_map_i[i].idx < NoMstPorts) else
-      $fatal(1, $sformatf("This rule has a slave id that is not allowed!!!\n\
+      $fatal(1, $sformatf("This rule has a IDX that is not allowed!!!\n\
           Violating rule %d.\n\
           Rule> IDX: %h START: %h END: %h\n\
           Rule> MAX_IDX: %h\n\
--- a/src/addr_decode.sv
+++ b/src/addr_decode.sv
@@ -10,26 +10,34 @@
 
 // Author: Wolfgang Roenninger <wroennin@ethz.ch>
 
-// Address Decoder: Maps the input address combinational to an index.
+// Address Decoder: Maps the input address combinatorially to an index.
 // The Address Map `addr_map_i` is a packed array of rule_t structs.
-// The rule on the MSB position in the array wins if there is an overlap
-// with the ranges.
-
+// The ranges of any two rules may overlap. If so, the rule at the higher (more significant)
+// position in `addr_map_i` prevails.
+//
 // The address decoder expects 3 fields in `rule_t`:
-//  - `idx`:        index of the rule, has to be < `NoIndices`, should be of type `int unsigned`
-//  - `start_addr`: start address of the range the rule describes, is included
-//  - `end_addr`:   end address of the range the rule describes, is NOT included
+//
+// typedef struct packed {
+//   int unsigned idx;
+//   addr_t       start_addr;
+//   addr_t       end_addr;
+// } rule_t;
+//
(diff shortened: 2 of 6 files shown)


src/addr_decode.sv
```diff
@@ -0,0 +1,125 @@
...
+
+// Address Decoder: Maps the input address combinational to an index.
+// The Address Map `addr_map_i` is a packed array of rule_t structs.
+// The rule on the MSB position in the array wins if there is an overlap
+// with the ranges.
+
+// The address decoder expects 3 fields in `rule_t`:
+//  - `idx`:        index of the rule, has to be < `NoIndices`, should be of type `int unsigned`
+//  - `start_addr`: start address of the range the rule describes, is included
+//  - `end_addr`:   end address of the range the rule describes, is NOT included
```
accuminium: I suggest your provide a template struct here with comments that describe the fields (the RHS of the `:` currently).

src/addr_decode.sv
```diff
@@ -0,0 +1,125 @@
...
+// `en_default_idx_i`: Driving this port to `1'b1` maps all input addresses
+// for which no rule in `addr_map_i` exists to the default index specified by
+// `default_idx_i`.  In this case, `dec_error_o` is always `1'b0`.
+
+// Assertions: The module checks every time there is a change in the address mapping
+// if the resulting map is valid. It fatals if `start_addr` is higher than `end_addr`
+// or if a mapping targets a port index that is outside the number of allowed indices.
+// It issues warnings if the address regions of any two mappings overlap.
+
+module addr_decode #(
```
accuminium: Could you please add a minimal testbench that checks the essentials?  I would consider:
- decoding an address (with a map of two or more entries) to the correct index
- decoding to the default index when no entry matches and the default is enabled
- raising a decode error when no entry matches and the default is disabled

src/addr_decode.sv
```diff
@@ -0,0 +1,125 @@
...
+// compliance with the License.  You may obtain a copy of the License at
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+// Author: Wolfgang Roenninger <wroennin@ethz.ch>
+
+// Address Decoder: Maps the input address combinational to an index.
```
accuminium:
```suggestion
// Address Decoder: Maps the input address combinatorially to an index.
```
An adverb is needed, and according to MW "combinationally" is not a word but ["combinatorially"](https://www.merriam-webster.com/dictionary/combinatorially) is.

src/addr_decode.sv
```diff
@@ -0,0 +1,125 @@
...
+// The Address Map `addr_map_i` is a packed array of rule_t structs.
+// The rule on the MSB position in the array wins if there is an overlap
+// with the ranges.
+
+// The address decoder expects 3 fields in `rule_t`:
+//  - `idx`:        index of the rule, has to be < `NoIndices`, should be of type `int unsigned`
+//  - `start_addr`: start address of the range the rule describes, is included
+//  - `end_addr`:   end address of the range the rule describes, is NOT included
+// There can be an arbitrary number of address rules. There can be multiple
+// ranges defined for the same index. The start address has to be <= the end address.
```
accuminium:
```suggestion
// ranges defined for the same index. The start address has to be less or equal than the end address.
```

src/addr_decode.sv
```diff
@@ -0,0 +1,125 @@
...
+// There isno default, to enable
```
accuminium: Please complete this sentence.

src/addr_decode.sv
```diff
@@ -0,0 +1,125 @@
...
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+// Author: Wolfgang Roenninger <wroennin@ethz.ch>
+
+// Address Decoder: Maps the input address combinational to an index.
+// The Address Map `addr_map_i` is a packed array of rule_t structs.
+// The rule on the MSB position in the array wins if there is an overlap
+// with the ranges.
```
accuminium: Suggest to rephrase as "The ranges of any two rules may overlap. If so, the rule at the higher (more significant) position in `addr_map_i` prevails."

README.md
```diff
@@ -53,6 +53,7 @@ Please note that cells with status *deprecated* are not to be used for new desig
 
 | Name                         | Description                                                                    | Status         | Superseded By |
 | :--------------------------- | :----------------------------------------------------------------------------- | :------------- | :------------ |
+| `addr_decode   `             | Address map decoder                                                    | active         |               |
```
accuminium: The alignment here seems to be broken.

src/addr_decode.sv
WRoenninger: changed it to `less than` as the comparison in the always comb block will not trigger if the start equals the end address, I could add there an additional check, would then however lead to more logic

src/addr_decode.sv
```diff
@@ -0,0 +1,133 @@
...
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+// Author: Wolfgang Roenninger <wroennin@ethz.ch>
+
+// Address Decoder: Maps the input address combinatorially to an index.
+// The Address Map `addr_map_i` is a packed array of rule_t structs.
```
zarubaf:
```suggestion
// The address map `addr_map_i` is a packed array of rule_t structs.
```

src/addr_decode.sv
```diff
@@ -0,0 +1,133 @@
...
+// The ranges of any two rules may overlap. If so, the rule at the higher (more significant)
+// position in `addr_map_i` prevails.
+//
+// The address decoder expects 3 fields in `rule_t`:
```
zarubaf:
```suggestion
// The address decoder expects three fields in `rule_t`:
```

src/addr_decode.sv
```diff
@@ -0,0 +1,133 @@
...
+//  - `start_addr`: start address of the range the rule describes, value is included in range
+//  - `end_addr`:   end address of the range the rule describes, value is NOT included in range
+//
+// There can be an arbitrary number of address rules. There can be multiple
+// ranges defined for the same index. The start address has to be less than the end address.
+//
+// There is the possibility to add a default mapping:
+// `en_default_idx_i`: Driving this port to `1'b1` maps all input addresses
+// for which no rule in `addr_map_i` exists to the default index specified by
+// `default_idx_i`.  In this case, `dec_error_o` is always `1'b0`.
```
zarubaf:
```suggestion
// `default_idx_i`. In this case, `dec_error_o` is always `1'b0`.
```

src/addr_decode.sv
```diff
@@ -0,0 +1,133 @@
...
+  // These following assumptions check the validity of the address map.
+  // The assumptions gets generated for each distinct pair of rules.
+  // Each assumption is present two times, as they rely on one rules being
+  // effectively ordered. Only one of the rules with the same function is
+  // active at a time for a given pair.
+  // check_start:        Enforces a smaller start than end address.
+  // check_idx:          Enforces a valid index in the rule.
+  // check_overlap:      Warns if there are overlapping address regions.
+  for (genvar i = 0; i < NoRules; i++) begin : gen_assert_0
+    check_start : assume final (addr_map_i[i].start_addr < addr_map_i[i].end_addr) else
```
zarubaf: Can we check that this assertion doesn't trigger in a simulation that does not fix the address ports to constants? I think we might want to change the checking of the address fields only to when the decoding is valid.

WRoenninger: Something like this?
```systemverilog
  always @(addr_map_i) #0 begin : proc_check_addr_map
    for (int unsigned i = 0; i < NoRules; i++) begin
      check_start : assume (addr_map_i[i].start_addr < addr_map_i[i].end_addr) else
        $fatal(1, $sformatf("This rule has a higher start than end address!!!\n\
            Violating rule %d.\n\
            Rule> IDX: %h START: %h END: %h\n\
            #####################################################",
            i ,addr_map_i[i].idx, addr_map_i[i].start_addr, addr_map_i[i].end_addr));
      // check the SLV ids
      check_idx : assume (addr_map_i[i].idx < NoIndices) else
          $fatal(1, $sformatf("This rule has a IDX that is not allowed!!!\n\
          Violating rule %d.\n\
          Rule> IDX: %h START: %h END: %h\n\
          Rule> MAX_IDX: %h\n\
          #####################################################",
          i, addr_map_i[i].idx, addr_map_i[i].start_addr, addr_map_i[i].end_addr,
          (NoIndices-1)));
      for (int unsigned j = i + 1; j < NoRules; j++) begin
        // overlap check
        check_overlap : assume (!((addr_map_i[j].start_addr < addr_map_i[i].end_addr) &&
                                  (addr_map_i[j].end_addr > addr_map_i[i].start_addr)))   else
             $warning($sformatf("Overlapping address region found!!!\n\
            Rule %d: IDX: %h START: %h END: %h\n\
            Rule %d: IDX: %h START: %h END: %h\n\
            #####################################################",
            i, addr_map_i[i].idx, addr_map_i[i].start_addr, addr_map_i[i].end_addr,
            j, addr_map_i[j].idx, addr_map_i[j].start_addr, addr_map_i[j].end_addr));
      end
    end
```

zarubaf: Hm, I would probably check on rising clock and disable iff the reset is asserted and the decoding is non valid.

WRoenninger: problem with that is that the module has no clock as it is fully combinational

src/addr_decode.sv
accuminium: Okay

PR #59, opened 2019-12-03
cdc_fifo_gray: Switch to a safe CDC FIFO design

zarubaf:
Largely follows these guidelines: http://www.sunburst-design.com/papers/CummingsSNUG2008Boston_CDC.pdf

### DIFF
--- a/src/cdc_fifo_gray.sv
+++ b/src/cdc_fifo_gray.sv
@@ -1,4 +1,4 @@
-// Copyright 2018 ETH Zurich and University of Bologna.
+// Copyright 2018-2019 ETH Zurich and University of Bologna.
 //
 // Copyright and related rights are licensed under the Solderpad Hardware
 // License, Version 0.51 (the "License"); you may not use this file except in
@@ -10,6 +10,9 @@
 // specific language governing permissions and limitations under the License.
 //
 // Fabian Schuiki <fschuiki@iis.ee.ethz.ch>
+// Florian Zaruba <zarubaf@iis.ee.ethz.ch>
+
+`include "common_cells/registers.svh"
 
 /// A clock domain crossing FIFO, using gray counters.
 ///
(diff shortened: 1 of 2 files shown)


src/cdc_fifo_gray.sv
```diff
@@ -42,117 +56,171 @@ module cdc_fifo_gray #(
...
+
+  // Cut the combinatorial path with a spill register.
+  spill_register #(
+    .T       ( T           )
+  ) i_spill_register (
+    .clk_i   ( dst_clk_i   ),
+    .rst_ni  ( dst_rst_ni  ),
+    .valid_i ( dst_valid   ),
+    .ready_o ( dst_ready   ),
+    .data_i  ( dst_data    ),
```
accuminium: Does the asynchronous data not have to be synchronized as well, like the read and write pointers?

zarubaf: No, we completely expose the data of the fifo to the receiving module. The read pointer indexing the content is synchronous to the `dst` domain. We just need to make sure that the written data is eventually communicated to the `dst` side and that the `dst` side has a pessimistic view on the content. We achieve this by
```
set_max_delay 500 \
     -through [get_pins -hierarchical -filter async] \
    -through [get_pins -hierarchical -filter async]
 set_false_path -hold \
     -through [get_pins -hierarchical -filter async] \
     -through [get_pins -hierarchical -filter async]
```
Furthermore, we provide at least two more cycles (synchronization of the pointers) which makes sure that we read the correct data (a.k.a the pessimistic view).

accuminium: Thanks. I understand that the read pointer is synchronized, but I was concerned about metastability on the data path, where `dst_data` is sampled by `dst_clk_i`, to which it is asynchronous. I think I now understand how the combination of `set_max_delay` and the synchronization of the read pointer ensures that no metastability on the data can occur: The delayed read pointer ensures that the data in the asynchronous FIFO is stable for some destination clock cycles, which, together with the max delay, fulfills the setup condition of the destination. The hold condition of the destination can be fulfilled with the delayed read pointer alone.

Is there no dependency between `SYNC_STAGES` and `LOG_DEPTH` for this to work properly, though? Something like
```
2^LOG_DEPTH >= 2*SYNC_STAGES
```

accuminium: I think adding a brief comment that describes this mechanism increases maintainability and confidence of/in this module.

zarubaf: Completely agree. I will add some more commentary in general about the module.

accuminium: Great, thanks!

PR #60, opened 2020-01-06
Update popcount to support input of width 1

b1f6c1c4:
This will be used by [Ariane/scoreboard](https://github.com/pulp-platform/ariane/blob/master/src/scoreboard.sv#L167-L172) to support `NR_COMMIT_PORTS == 1`.


zarubaf:
Thanks!

PR #72, opened 2020-03-14
Add ECC module

zarubaf:
Earlier last year I was experimenting a bit with ECC. I thought this module might be interesting for somebody (e.g. error correction on an L2 for example). This module can detect up to two errors and correct a single error.

As the testbench is built using Verilator and C++ I also added the tests to CI. 

Happy to have your input on this.

### DIFF
--- a/src/ecc_pkg.sv
+++ b/src/ecc_pkg.sv
@@ -1,0 +1,31 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
+// Copyright and related rights are licensed under the Solderpad Hardware
+// License, Version 0.51 (the "License"); you may not use this file except in
+// compliance with the License. You may obtain a copy of the License at
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+//
+// Author: Florian Zaruba <zarubaf@iis.ee.ethz.ch>
+//
+// Description: Contains common ECC definitions and helper functions.
+
+package ecc_pkg;
+
+  // Calculate required ECC parity width:
+  function automatic int unsigned get_parity_width (input int unsigned data_width);
+    // data_width + cw_width + 1 <= 2**cw_width
+    int unsigned cw_width = 2;
+    while (unsigned'(2**cw_width) < cw_width + data_width + 1) cw_width++;
+    return cw_width;
+  endfunction
+
+  // Calculate required ECC codeword width:
+  function automatic int unsigned get_cw_width (input int unsigned data_width);
+    // data width + parity width + one additional parity bit (for double error detection)
+    return data_width + get_parity_width(data_width);
+  endfunction
+
+endpackage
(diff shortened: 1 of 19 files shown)


test/ecc/ecc_decode.cpp
```diff
@@ -0,0 +1,114 @@
...
+
+        if (main_time > 100000) break;
+    }
+    dut->final();
+    tfp->close();
+    delete dut;
+    exit(0);
+}
+
+
```
accuminium:
```suggestion
```

test/ecc/ecc_encode.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+        if (main_time > 100000) break;
+    }
+    dut->final();
+    tfp->close();
+    delete dut;
+    exit(0);
+}
+
+
```
accuminium:
```suggestion
```

test/ecc/ecc_decode.cpp
```diff
@@ -0,0 +1,114 @@
...
+        single_error = 0;
+        double_error = 0;
+        parity_error = 0;
+
+        switch (random() % 10) {
+            // inject single error
+            case 0:
+                printf("Injecting single error\n");
+                single_error = 1;
+                dut->data_i[0] ^= 0x4;
```
accuminium: Would it be feasible to increase the coverage of error correction (and error detection in case 2 below) by randomizing the position of the injected error?

test/ecc_synth.sv
```diff
@@ -0,0 +1,26 @@
...
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+module ecc_synth;
+
+    for (genvar dw = 8; dw < 32; dw += 8) begin
+        logic [ecc_pkg::get_cw_width(dw):0] codeword;
```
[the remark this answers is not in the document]
zarubaf: That's the codeword + parity bit (which is technically not part of the codeword) so we track it separately. I have added a comment. Do you think we should be more verbose about it?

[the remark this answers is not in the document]
accuminium: Ah, I see. How about increasing the return value in `get_cw_width` by one?

zarubaf:
Thanks for the extensive input. CI needed some fixing (manual install a newer verilator version) due to the "new" SV features used.

zarubaf:
@accuminium @fabianschuiki  Is the PR okay? Any further changes requested?

src/ecc_pkg.sv
```diff
@@ -24,7 +24,8 @@ package ecc_pkg;
 
   // Calculate required ECC codeword width:
   function automatic int unsigned get_cw_width (input int unsigned data_width);
-    return data_width + get_parity_width(data_width);
+    // data width + parity width + one additional parity bit (for double error detection)
+    return data_width + get_parity_width(data_width) + 1;
```
accuminium:
```suggestion
    return data_width + get_parity_width(data_width);
```
With the new `encoded_data_t`, this `+1` is redundant, no? Both times you call `get_cw_width`
https://github.com/pulp-platform/common_cells/blob/9d2b7c9c6f4dadd0e5f10a811ed3fac959c458be/src/ecc_decode.sv#L34
https://github.com/pulp-platform/common_cells/blob/9d2b7c9c6f4dadd0e5f10a811ed3fac959c458be/src/ecc_encode.sv#L27
you subtract `1` again.

zarubaf: Yes, you are right. Thanks!

zarubaf:
As I got your LGTM, I am going ahead with merging this.

PR #79, opened 2020-04-20
assertions: Add assertion header

stmach:
Analogously to the register macro definitions in include header
`registers.svh`, this commit adds macro definitions for assertions.

### DIFF
--- a/include/common_cells/assertions.svh
+++ b/include/common_cells/assertions.svh
@@ -1,0 +1,113 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
+//
+// Copyright and related rights are licensed under the Solderpad Hardware
+// License, Version 0.51 (the "License"); you may not use this file except in
+// compliance with the License. You may obtain a copy of the License at
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+// Author: Stefan Mach <smach@iis.ee.ethz.ch>
+
+// Common assertion and coverage defines for RTL designs
+`ifndef COMMON_CELLS_ASSERTIONS_SVH_
+`define COMMON_CELLS_ASSERTIONS_SVH_
+
+// Abridged Summary of available assertion and coverage macros:
+// `ASSERT:         immediate
+// `ASSERT_ERR:     immediate with custom error message
+// `ASSERTC:        concurrent clocked with implicit clock and reset
+// `ASSERTC_ERR:    concurrent clocked with implicit clock and reset and custom error message
+// `ASSERTC_CR:     concurrent clocked
+// `ASSERTC_CRERR:  concurrent clocked with custom error message
+
+
+// Immediate assertion
+// __cond: Boolean expression that is asserted
+// __label: Label to give this assertion
+`define ASSERT(__cond, __label) \
+  /``* pragma translate_off *``/ \
+`ifndef VERILATOR \
+  initial begin \
+    __label : assert (__cond) else $error(`"__cond`"); \
+  end \
+`endif \
+  /``* pragma translate_on *``/
+
+// Immediate assertion with custom error message
+// __cond: Boolean expression that is asserted
+// __err: String containing the error message
+// __label: Label to give this assertion
+`define ASSERT_ERR(__cond, __err, __label) \
+  /``* pragma translate_off *``/ \
+`ifndef VERILATOR \
+  initial begin \
+    __label : assert (__cond) else $error(`"__err`"); \
+  end \
+`endif \
+  /``* pragma translate_on *``/
+
+// Concurrent clocked assertion with implicit clock and reset
+// __prop: Property expression
+// __label: Label to give this assertion
+// Implicit:
+// clk_i: clock signal
+// rst_ni: reset signal (active low)
+`define ASSERTC(__prop, __label) \
+  /``* pragma translate_off *``/ \
+`ifndef VERILATOR \
+  __label : assert property (@(posedge clk_i) disable iff (!rst_ni) \
+                             __prop) \
+      else $error(`"__prop`"); \
+`endif \
+  /``* pragma translate_on *``/
+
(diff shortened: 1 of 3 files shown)


include/common_cells/assertions.svh
```diff
@@ -0,0 +1,181 @@
...
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+// Author: Stefan Mach <smach@iis.ee.ethz.ch>
+
+// Common assertion and coverage defines for RTL designs
+`ifndef ASSERTION_DEFINES_H_
+`define ASSERTION_DEFINES_H_
```
zarubaf, line 16: Complete nitpick but I think the include guard should be named like the file :D
```suggestion
`ifndef ASSERTIONS_DEFINES_H_
`define ASSERTIONS_DEFINES_H_
```

include/common_cells/assertions.svh
```diff
@@ -0,0 +1,181 @@
...
+// `ASSERT:         immediate
+// `ASSERT_ERR:     immediate with custom error message
+// `ASSERTC:        concurrent clocked with implicit clock and reset
+// `ASSERTC_ERR:    concurrent clocked with implicit clock and reset and custom error message
+// `ASSERTC_CR:     concurrent clocked
+// `ASSERTC_CRERR:  concurrent clocked with custom error message
+// `ASSERTCI:       concurrent clocked using implication with implicit clock and reset
+// `ASSERTCI_ERR:   concurrent clocked using implication with implicit clock and reset w/ error msg.
+// `ASSERTCI_CR:    concurrent clocked using implication
+// `ASSERTCI_CRERR: concurrent clocked using implication with custom error message
```
zarubaf, line 28: Do you think those make sense? I have the feeling that the implicit implication operator actually inhibits readability. Feel free to start a discussion :D

include/common_cells/assertions.svh
accuminium: I suggest `COMMON_CELLS_ASSERTIONS_SVH_` in line with [`registers.svh`](https://github.com/pulp-platform/common_cells/blob/master/include/common_cells/registers.svh) and to minimize potential conflicts.

accuminium:
> We probably need to adapt:
> 
>     * `Bender.yml`
> 
>     * IPApprox file

Those do not need to be changed because only the include directory is specified.

include/common_cells/assertions.svh
stmach: You're right, this was even an old file name. `registers.svh` is using the following include guard: `COMMON_CELLS_REGISTERS_SVH_` so I shall change this one to follow the same structure.

stmach:
> 
> 
> > We probably need to adapt:
> > ```
> > * `Bender.yml`
> > 
> > * IPApprox file
> > ```
> 
> Those do not need to be changed because only the include directory is specified.

those are actually the only ones I checked and didn't think of the README 😅. Will update the missing ones.

zarubaf:
> > We probably need to adapt:
> > ```
> > * `Bender.yml`
> > 
> > * IPApprox file
> > ```
> 
> Those do not need to be changed because only the include directory is specified.

Ah perfect, didn't think of that.

include/common_cells/assertions.svh
zarubaf: Agree. That makes a lot of sense.

include/common_cells/assertions.svh
stmach: While I haven't really used implications in assertions much, I included them as the implication is necessary to assert properties with sequences. I don't know what the usual worklflow is for these kinds of properties.

A clean way may be to specify separate `sequence` blocks and create the assertion using these (using the `` `ASSERTCI`` macros), or create a `property` block containing sequences first and then assert that (using the non-implicative `` `ASSERT`` / `` `ASSERTC``).

zarubaf: But isn't the current implication macro just syntactic sugar that solves typing `|->`. You could also just specify this in the property for the concurrent assertion, right?

stmach: Yes, that's true. You're right that having twice as many macros with more arguments probably makes it less readable than actually seeing the `->` if the assertion is written directly into the macro.

stmach:
I removed the implication macros and added respecitve notes into `README` and `CHANGELOG`.

zarubaf:
Anything you want to add @accuminium?

include/common_cells/assertions.svh
```diff
@@ -0,0 +1,113 @@
...
+
+// Immediate assertion with custom error message
+// __cond: Boolean expression that is asserted
+// __err: String containing the error message
+// __label: Label to give this assertion
+`define ASSERT_ERR(__cond, __err, __label) \
+  /``* pragma translate_off *``/ \
+`ifndef VERILATOR \
+  initial begin \
+    __label : assert (__cond) else $error(`"__err`"); \
```
zarubaf, line 47: I was just using this macro and I think we should remove the ``"` for the error message as (at least VCS) complains about having double colons if the macro is used like this:
```verilog
  `ASSERT_ERR(Cfg.CacheLineWidth == Cfg.AxiDataWidth == Cfg.AxiSnpDataWidth, "Cache line width, AXI data width and Snoop address width need to be the same.", dev_constraints)
```
I think we could use it w/o `"` but that looks strange and as soon as we encounter the first comma this will get messy :D  @stmach What's your point on this?

stmach: Sure, makes sense. I think my idea was to "stringify" everything that's passed to the macro but i see the issue with string literals. I'm not sure removing the quotes works if you have a string coming from another macro.

Just to confirm on your second point, does it work when you remove the `` `"`` from the macro and pass the error message as a string literal? In that case I would say we just remove the quotes from the macro.

zarubaf:
> LGTM and definitely useful, thanks!
> 
> How does this compare to https://github.com/lowRISC/opentitan/blob/master/hw/ip/prim/rtl/prim_assert.sv, which is mentioned in the Coding Conventions we follow and on which we might want to converge at some point in the future?

I think that is exactly what we need, from my POV there is no need to re-invent. Should we just update on the lowrisc defines?

zarubaf:
I have updated the assertions header with the one from lowrisc. Please have a look at whether that is suitable for your needs or let me know if I should change anything.

stmach (commented):
LGTM

PR #81, opened 2020-04-28
Throughput fairness for rr_arb_tree

WRoenninger:
The module `rr_arb_tree` does not distribute the throughput evenly when not all input requests are active.

This adds additional logic, which jumps the internal round robin state to the next active input request in respect to the currently served one.
In addition there is now a dedicated testbench for `rr_arb_tree`, which checks for the throughput and validity of arbitrated data.

### DIFF
--- a/src/rr_arb_tree.sv
+++ b/src/rr_arb_tree.sv
@@ -15,11 +15,11 @@
 /// The rr_arb_tree employs non starving round robin arbitration - i.e. the priorities
 /// rotate each cycle.
 module rr_arb_tree #(
-  /// Number of handshaked inputs which get arbitrated.
+  /// Number of inputs to be arbitrated.
   parameter int unsigned NumIn      = 64,
   /// Data width of the payload in bits. Not needed if `DataType` is overwritten.
   parameter int unsigned DataWidth  = 32,
-  /// Data type of the payload, can be overwritten with custom type.
+  /// Data type of the payload, can be overwritten with custom type. Only use of `DataWidth`.
   parameter type         DataType   = logic [DataWidth-1:0],
   /// The `ExtPrio` option allows to override the internal round robin counter via the
   /// `rr_i` signal. This can be useful in case multiple arbiters need to have
@@ -41,23 +41,38 @@ module rr_arb_tree #(
   parameter bit          LockIn     = 1'b0,
   /// When set, ensures that throughput gets distributed evenly between all inputs.
   /// Set to `1'b0` to disable.
-  parameter bit          FairArb    = 1'b1
+  parameter bit          FairArb    = 1'b1,
+  /// Dependent parameter, do **not** overwrite.
+  /// Width of the arbitration priority signal and the arbitrated index.
+  parameter int unsigned IdxWidth   = (NumIn > 32'd1) ? unsigned'($clog2(NumIn)) : 32'd1,
+  /// Dependent parameter, do **not** overwrite.
+  /// Type for defining the arbitration priority and arbitrated index signal.
+  parameter type         idx_t      = logic [IdxWidth-1:0]
 ) (
-  input  logic                             clk_i,
-  input  logic                             rst_ni,
-  input  logic                             flush_i, // clears the arbiter state
-  input  logic [$clog2(NumIn)-1:0]         rr_i,    // external RR prio (needs to be enabled above)
-  // input requests and data
-  input  logic [NumIn-1:0]                 req_i,
+  /// Clock, positive edge triggered.
+  input  logic                clk_i,
+  /// Asynchronous reset, active low.
+  input  logic                rst_ni,
+  /// Clears the arbiter state. Only used if `ExtPrio` is `1'b0` or `LockIn` is `1'b1`.
+  input  logic                flush_i,
+  /// External round-robin priority. Only used if `ExtPrio` is `1'b1.`
+  input  idx_t                rr_i,
+  /// Input requests arbitration.
+  input  logic    [NumIn-1:0] req_i,
   /* verilator lint_off UNOPTFLAT */
-  output logic [NumIn-1:0]                 gnt_o,
+  /// Input request is granted.
+  output logic    [NumIn-1:0] gnt_o,
   /* verilator lint_on UNOPTFLAT */
-  input  DataType [NumIn-1:0]              data_i,
-  // arbitrated output
-  input  logic                             gnt_i,
-  output logic                             req_o,
-  output DataType                          data_o,
-  output logic [$clog2(NumIn)-1:0]         idx_o
+  /// Input data for arbitration.
+  input  DataType [NumIn-1:0] data_i,
+  /// Output request is valid.
+  output logic                req_o,
+  /// Output request is granted.
+  input  logic                gnt_i,
+  /// Output data.
+  output DataType             data_o,
+  /// Index from which input the data came from.
+  output idx_t                idx_o
 );
 
   // pragma translate_off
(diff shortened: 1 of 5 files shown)


zarubaf:
Thanks for bringing this up. Do you have an estimate on area/timing between fair and unfair arbitration?

WRoenninger:
> Thanks for bringing this up. Do you have an estimate on area/timing between fair and unfair arbitration?

Not at the moment. I can do some synthesis in GF22, however are not really sure how I should constrain the arbiter.
The additional logic has a leading_zero_conter and needs the arbitration decision (`idx_o`) as input.
Which will probably lead to a longer internal path to the arbitration priority register.

WRoenninger:
I did some synthesis and the general behavior is that the more `NumIn` the design has, the more the impact of this change is can be felt.
In gf22 all reg2reg delays where longer (~2x), however should allow for 1 GHz constraints. 
Inp2reg timings are longer (~2x). Critical inputs are coming from `req_i`.
The main contribution to the delays is the `lzc` deployed to calculate the jump.
Area wise the impact is also determined by the number of inputs,  is however max +5% and is less the wider the `DataWidth` is.

zarubaf:
Thanks for conducting these tests. My general feeling ist that we should probably take more time to investigate if we can reduce the timing pressure by re-architecting the arbiter. Given that this design is at the foundation of all our interconnect IPs a `2x` increase is quite significant.

WRoenninger:
As I see it is the main reason for the increase in delay the dependency of `mask_req` on the signal `idx_o`. And the longest path has to basically traverse two logarithmic trees.

WRoenninger:
Now the next arbiter state for fair throughput only depends on `req_i` and `rr_q`.
Synthesis shows:
- Marginal increase in size as adder for the next state is replaced by two `lzc`.
- Timings:
  - `in2out`: no change to unfair arb
  - `reg2out`: no change to unfair arb
  - `reg2reg`:  goes from `rr_q` over two stages into an `lzc` to `rr_d`, was from `lock_d` to `lock_q`
  - `in2reg`: goes from `req_i` over an `lzc` to `reg_q` (is longer than previously) as it only went 
    from `req_i` to `lock_d`.

src/rr_arb_tree.sv
```diff
@@ -142,8 +163,42 @@ module rr_arb_tree #(
         assign req_d = req_i;
       end
 
-      assign rr_d       = (gnt_i && req_o) ? ((rr_q == NumLevels'(NumIn-1)) ? '0 : rr_q + 1'b1) : rr_q;
+      if (FairArb) begin : gen_fair_arb
```
zarubaf, line 166: Can you please add a description of the architecture and why we opted for two LZC? Just so that we remember this in a couple of weeks :)

test/rr_arb_tree_tb.sv
```diff
@@ -0,0 +1,266 @@
...
+// compliance with the License.  You may obtain a copy of the License at
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+// Author: Wolfgang Roennigner <wroennin@iis.ee.ethz.ch>
+
+timeunit 1ns/1ps;
```
zarubaf, line 13:
```suggestion
```

test/rr_arb_tree_tb.sv
```diff
@@ -0,0 +1,266 @@
...
+        @(posedge clk);
+      end
+
+      for (int unsigned j = 1; j <= NumInp; j++) begin
+        if (tot_active[j] > 0) begin
+          throughput  = real'(tot_served[j])/real'(tot_active[j]);
+          exp_through = real'(1)/real'(j);
+          error       = throughput - exp_through;
+          if (FairArb && LockIn) begin
+            assert(error < error_threshold && error > -error_threshold);
```
zarubaf, line 215: Nice! Just would add a comment in the assertion in case it fires. Like `Arbiter not fair.` or something alike.

src/rr_arb_tree.sv
```diff
@@ -11,49 +11,68 @@
...
-// is needed, just connect `rr_i` to '0.
-//
-// If `AxiVldRdy` is set, the req/gnt signals are compliant with the AXI style vld/rdy
-// handshake. Namely, upstream vld (req) must not depend on rdy (gnt), as it can be deasserted
-// again even though vld is asserted. Enabling `AxiVldRdy` leads to a reduction of arbiter
-// delay and area.
-//
 
+/// The rr_arb_tree employs non starving round robin arbitration - i.e. the priorities
+/// rotate each cycle.
```
zarubaf, line 16: I think we should add the discussion from this PR on how the critical path and behavior change when fair arbitration is activated. Maybe make a new heading `## Fair vs. Unfair Arbitration`.

WRoenninger: Added fairness points to the module inline description.

src/rr_arb_tree.sv
```diff
@@ -11,49 +11,102 @@
...
+/// ## Fair vs. unfair Arbitration
+///
+/// This refers to fair throughput distribution when not all inputs have active requests.
+/// This module has an internal state `rr_q` which defines the highest priority input. (When
+/// `ExtPrio` is `1'b1` this state is provided from the outside.) The arbitration tree will
+/// choose the input with the same index as currently defined by the state if it has an active
+/// request. Otherwise a *random* other active input is selected. The parameter `FairArb` is used
+/// to distinguish between two methods of calculating the next state.
+/// * `1'b0`: The next state is calculated by advancing the current state by one. This leads to the
+///           state be being calculated without context of the active request. Leading to
```
zarubaf, line 27:
```suggestion
///           state being calculated without the context of the active request. Leading to an
```

src/rr_arb_tree.sv
```diff
@@ -11,49 +11,102 @@
...
+///           unfair throughput distribution if not all inputs have active requests.
+/// * `1'b1`: The next state jumps to the next unserved request with higher index.
+///           This is achieved by using two trailing-zero-counters (`lzc`). The upper has the masked
+///           `req_i` signal with all indicies which will have a higher priority in the next state.
```
zarubaf, line 31:
```suggestion
///           `req_i` signal with all indices which will have a higher priority in the next state.
```

src/rr_arb_tree.sv
```diff
@@ -11,49 +11,102 @@
...
+///           The trailing zero count defines the input index with the next highest priority after
+///           the current one is served. When the upper is empty the lower `lzc` provides the
+///           wrapped index if there are outstanding requests with lower or same priority.
+/// The implication of throughput fairness on the module timing are:
+/// * The trailing zero counter (`lzc`) has a loglog relation of input to output timing. This means
+///   that in this module the input to register path scales also with Log(Log(O)) in relation to
```
zarubaf, line 37: I think the `O` can easily be misread as a `0`
```suggestion
///   that in this module the input to register path scales also with Log(Log(NrOup)) in relation to
```

src/rr_arb_tree.sv
```diff
@@ -11,49 +11,102 @@
...
+///   the number of inputs.
+/// * The `rr_arb_tree` data multiplexing scales with Log(O). This means that the input to output
+///   timing path of this module also scales scales with Log(O).
+/// This implies that in this module the input to output path is always longer than the input to
+/// register path. As the output data usually also terminates in a register the parameter `FairArb`
+/// only has implications on the area. When it is `1'b0`a static plus one adder is instantiated.
```
zarubaf, line 43:
```suggestion
/// only has implications on the area. When it is `1'b0` a static plus one adder is instantiated.
```

src/rr_arb_tree.sv
```diff
@@ -11,49 +11,102 @@
...
+/// If it is `1'b1` two `lzc`, a masking logic stage and a two input multiplexer are instantiated.
+/// However these are small in respect of the data multiplexers needed, as the width of the `req_i`
+/// signal is usually less as the `DataWidth`.
```
zarubaf, line 46:
```suggestion
/// signal is usually less than the `DataWidth`.
```

src/rr_arb_tree.sv
WRoenninger: Changed it to `NrInp` which is the parameter it scales from.

WRoenninger:
Added the suggestions.

zarubaf:
Merged into master.

PR #83, opened 2020-05-10
Isochronous spill register

zarubaf:
This module works similarly to the `spill_register` as it cuts any combinatorial paths but additionally allows to couple two clock domains with an isochronous clock (i.e., two clocks which do not vary in phase but vary in a constant integer factor in frequency).

The implementation concept is similar to the CDC FIFO but without any synchronization mechanisms as we do not need to deal with metastability (timing covered by STA).

### DIFF
--- a/src/isochronous_spill_register.sv
+++ b/src/isochronous_spill_register.sv
@@ -1,0 +1,87 @@
+// Copyright 2020 ETH Zurich and University of Bologna.
+//
+// Copyright and related rights are licensed under the Solderpad Hardware
+// License, Version 0.51 (the "License"); you may not use this file except in
+// compliance with the License. You may obtain a copy of the License at
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+//
+// Florian Zaruba <zarubaf@iis.ee.ethz.ch>
+
+`include "common_cells/registers.svh"
+
+/// A register with handshakes that completely cuts any combinatorial paths
+/// between the input and output in isochronous clock domains.
+///
+/// > Definition of isochronous: In telecommunication, an isochronous signal is a signal
+/// > in which the time interval separating any two significant instants is equal to the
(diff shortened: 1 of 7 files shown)


src/isochronous_spill_register.sv
```diff
@@ -0,0 +1,87 @@
...
+/// > in which the time interval separating any two significant instants is equal to the
+/// > unit interval or a multiple of the unit interval.
+///
+/// The source and destination clock domains must be derived from the same clock
+/// but can vary in frequency by a constant factor (e.g., double the frequency).
+///
+/// The module is basically a two deep dual-clock fifo with read and write pointers
+/// in different clock domains. As we know the static timing relationship between the
+/// clock domains we can rely on static timing analysis (STA) to get the sampling windows
+/// right and therefore don't need any synchronization.
```
fabianschuiki, line 29: This is just beautiful :+1:

src/isochronous_spill_register.sv
```diff
@@ -0,0 +1,87 @@
...
+    `FFLNR(mem_q, mem_d, (src_valid_i && src_ready_o), src_clk_i)
+    always_comb begin
+      mem_d = mem_q;
+      mem_d[wr_pointer_q[0]] = src_data_i;
+    end
+
+    assign src_ready_o = (rd_pointer_q ^ wr_pointer_q) != 2'b10;
+
+    assign dst_valid_o = (rd_pointer_q ^ wr_pointer_q) != '0;
+    assign dst_data_o = mem_q[rd_pointer_q[0]];
```
fabianschuiki, line 85: Describing this like a FIFO is actually a very elegant solution. I like it a lot. Much better than the baseline spill register, which has a rather weird setup.

zarubaf:
> Looking at the isochronous spill register, I wonder what's actually different from the regular spill register. If they are essentially the same, we could just replace the implementation of `spill_register` with this one, and have a module that works for both cases!

I was thinking about that as well. There is one slight difference that I can see. The current `spill_register` needs one additional state-holding bit per slice (`valid`/`full`) while the FIFO variant would need two (the two bits per pointer). As this is a quite widely used IP throughout the project I have a tendency to keep the `spill_register` as the optimized version as it is now. What do you think?

fabianschuiki:
> As this is a quite widely used IP throughout the project I have a tendency to keep the `spill_register` as the optimized version as it is now. What do you think?

Yes I very much agree with that. In that case, let's keep the two!

PR #88, opened 2020-05-20
Index width constant function.

WRoenninger:
This adds a the function `cf_math_pkg::idx_width()`.

* Many modules have special degeneration code for `$clog2(x)` where `x==1`.   
* Adds a constant function which ensures that defined index signals always have a width of 1.
* Added the function to `addr_decode` and `lzc` as example how to use.
* Changed `addr_decode` and `lzc` to inline documentation.

### DIFF
--- a/src/addr_decode.sv
+++ b/src/addr_decode.sv
@@ -15,18 +15,6 @@
 /// The ranges of any two rules may overlap. If so, the rule at the higher (more significant)
 /// position in `addr_map_i` prevails.
 ///
-/// The address decoder expects three fields in `rule_t`:
-///
-/// typedef struct packed {
-///   int unsigned idx;
-///   addr_t       start_addr;
-///   addr_t       end_addr;
-/// } rule_t;
-///
-///  - `idx`:        index of the rule, has to be < `NoIndices`
-///  - `start_addr`: start address of the range the rule describes, value is included in range
-///  - `end_addr`:   end address of the range the rule describes, value is NOT included in range
-///
 /// There can be an arbitrary number of address rules. There can be multiple
 /// ranges defined for the same index. The start address has to be less than the end address.
 ///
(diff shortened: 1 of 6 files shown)


src/addr_decode.sv
```diff
@@ -10,53 +10,73 @@
...
 ) (
-  input  addr_t               addr_i,           // address to decode
-  input  rule_t [NoRules-1:0] addr_map_i,       // address map: rule with the highest position wins
-  output idx_t                idx_o,            // decoded index
-  output logic                dec_valid_o,      // decode is valid
-  output logic                dec_error_o,      // decode is not valid
-  // Default index mapping enable
-  input  logic                en_default_idx_i, // enable default port mapping
-  input  idx_t                default_idx_i     // default port index
+  /// Address to decode
```
zarubaf, line 60:
```suggestion
  /// Address to decode.
```

src/addr_decode.sv
```diff
@@ -10,53 +10,73 @@
...
+  input  addr_t               addr_i,
+  /// Address map: rule with the highest array position wins on collision
```
zarubaf, line 62: I think it would make sense to move the struct definition here:
```suggestion
  /// Address map: rule with the highest array position wins on a collision
  /// ```
  /// typedef struct packed {
  ///   int unsigned idx;
  ///   addr_t       start_addr;
  ///   addr_t       end_addr;
  /// } rule_t;
  ///```
```

src/addr_decode.sv
```diff
@@ -10,53 +10,73 @@
...
+  input  rule_t [NoRules-1:0] addr_map_i,
+  /// Decoded index
```
zarubaf, line 64:
```suggestion
  /// Decoded index.
```

src/addr_decode.sv
```diff
@@ -10,53 +10,73 @@
...
+  output idx_t                idx_o,
+  /// Decode is valid
```
zarubaf, line 66:
```suggestion
  /// Decode is valid.
```

src/addr_decode.sv
```diff
@@ -10,53 +10,73 @@
...
+  output logic                dec_valid_o,
+  /// Decode is not valid, no matching rule found
```
zarubaf, line 68:
```suggestion
  /// Decode is not valid, no matching rule found.
```

src/addr_decode.sv
WRoenninger: Moved it to the `rule_t` parameter definition.

zarubaf:
Merged

PR #89, opened 2020-06-16
src_files.yml addr_decode depends on cf_math

uge:
Fixes parsing order of files as addr_decode.sv depends on cf_math_pkg.sv


zarubaf:
Thanks, LGTM

PR #95, opened 2020-07-31
Add barrel shifter module

WRoenninger:
This adds a logarithmic barrel shifter.
* Can have different byte_width.
* Left or right parametrisation.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -7,7 +7,8 @@ package:
     - "Andreas Kurth <akurth@iis.ee.ethz.ch>"
     - "Manuel Eggimann <meggimann@iis.ee.ethz.ch>"
     - "Stefan Mach <smach@iis.ee.ethz.ch>"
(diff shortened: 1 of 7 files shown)


zarubaf:
I let Steve weigh in. I am not entirely convinced that this performers better than the Synthesiser's `<<` implementation. Do you have any numbers?

WRoenninger:
There where some numbers done by @thommythomaso. It seems that the design-ware implementation performs better.

stmach:
Unless there is a tangible case where this is required and better over `result = {input, input} >> shift` or similar, I don't really see the necessity for a dedicated shifter module.
Realistically, any synthesizer will have a more or less target-optimized version of a shifter inferred from `>>`, which is also most portable.

zarubaf:
So I am closing this PR for now. In case there is a necessity for merging or other opinions feel free to re-open or create another PR.

PR #98, opened 2020-09-04
lint: Clean lint style fixes

zarubaf:
Running the source code through `verible-lint` revealed a couple
of lint style problems which are fixed by this commit.


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Fixed
 - `stream_register`: Fix `DATA_WIDTH` of instantiated FIFO.
 - `stream_xbar`: Add missing argument in assertion error string.
+- Lint style fixes
```
accuminium, line 15: I suggest you add
- deprecated `sram`
- added `stream_fork_seq`

to the Changelog.

Bender.yml
```diff
@@ -21,6 +21,8 @@ sources:
   # Files in level 1 only depend on files in level 0, files in level 2 on files in levels 1 and 0,
   # etc. Files within a level are ordered alphabetically.
   # Level 0
```
accuminium, line 23: `src/stream_fork_seq.sv` needs to be added

accuminium (approved):
The style cleanup looks good to me, thanks for the effort!

I have two points, but I think they require no second review. You might want to put them into separate commits.

accuminium:
Two other points:
1) Which style file did you use for linting?
2) How much effort would it be to add a linter check to GitHub CI?

CHANGELOG.md
zarubaf: Thanks a lot, I absolutely agree. I didn't want to introduce `stream_fork_seq`. Thanks for catching that. Absolutely agree regarding the `sram`.

zarubaf:
1. It is the default style guide from `verible` which is the lowrisc (and hence ours). I think you can do some customization to make it fit with other style guides.
2. Absolutely easy. That was my intention to add it to the CI after this commit, I already have it working for another project and it uses GitHub actions so it can be trivially copied to other repositories (for example the `axi`).

PR #101, opened 2020-10-06
add flushable spill_register

da-gazzi:
added a flushable spill_register and removed unused signals from the old spill_register.

### DIFF
--- a/src/spill_register.sv
+++ b/src/spill_register.sv
@@ -16,8 +16,7 @@
 /// between the input and output.
 module spill_register #(
   parameter type T      = logic,
-  parameter bit  Bypass = 1'b0,     // make this spill register transparent
-  parameter bit  EnableFlush = 1'b0 // enable flushing functionality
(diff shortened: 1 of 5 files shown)


src/spill_register.sv
```diff
@@ -37,7 +37,6 @@ module spill_register #(
     T a_data_q;
     logic a_full_q;
     logic a_fill, a_drain;
-    logic a_en, a_en_data;
 
     always_ff @(posedge clk_i or negedge rst_ni) begin : ps_a_data
       if (!rst_ni)
```
fabianschuiki, line 42: How about we instantiate `spill_register_v2` here, but with the `flush_i` signal tied to `0`?

fabianschuiki (changes requested):

da-gazzi:
I made `spill_register` a wrapper of `spill_register_flushable` as suggested by @fabianschuiki and fixed the dependency ordering :)

zarubaf:
Closed because of minor fixes in #119

PR #103, opened 2020-11-01
src_files: Change compile order

bluewww:
addr_decode depends on cf_math_pkg

### DIFF
--- a/src_files.yml
+++ b/src_files.yml
@@ -6,9 +6,8 @@ common_cells_all:
     # package. Files in level 1 only depend on files in level 0, files in level 2 on files in
     # levels 1 and 0, etc. Files within a level are ordered alphabetically.
     # Level 0
-    - src/cf_math_pkg.sv
-    - src/addr_decode.sv
     - src/cdc_2phase.sv
+    - src/cf_math_pkg.sv
     - src/clk_div.sv
     - src/delta_counter.sv
     - src/ecc_pkg.sv
@@ -38,6 +37,7 @@ common_cells_all:
     - src/sync_wedge.sv
     - src/unread.sv
     # Level 1
+    - src/addr_decode.sv
     - src/cb_filter.sv
     - src/cdc_fifo_2phase.sv
     - src/cdc_fifo_gray.sv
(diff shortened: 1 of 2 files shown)


accuminium:
Thanks for reporting this problem, @bluewww.  I can confirm [this dependency](https://github.com/pulp-platform/common_cells/blob/ff721776a7deabb0d09046c3aa8411d6e0686f63/src/addr_decode.sv#L53).

As a solution, why do you not move `addr_decode` to the "Level 1" modules (which by definition depend on "Level 0" modules)? That's the way it is [currently specified in `Bender.yml`](https://github.com/pulp-platform/common_cells/blob/ff721776a7deabb0d09046c3aa8411d6e0686f63/Bender.yml#L59).

bluewww:
Ok will do so

Bender.yml
```diff
@@ -21,8 +21,8 @@ sources:
   # Files in level 1 only depend on files in level 0, files in level 2 on files in levels 1 and 0,
   # etc. Files within a level are ordered alphabetically.
   # Level 0
-  - src/cdc_2phase.sv
   - src/cf_math_pkg.sv
+  - src/cdc_2phase.sv
```
zarubaf, line 25: Is there a reason why you moved `cdc_2phase`? We generally keep an alphabetical order amongst dependencies.

src_files.yml
```diff
@@ -6,9 +6,8 @@ common_cells_all:
     # package. Files in level 1 only depend on files in level 0, files in level 2 on files in
     # levels 1 and 0, etc. Files within a level are ordered alphabetically.
     # Level 0
-    - src/addr_decode.sv
-    - src/cdc_2phase.sv
     - src/cf_math_pkg.sv
+    - src/cdc_2phase.sv
```
zarubaf, line 10: Same here.

bluewww: ok I undo this. I just believe pkgs should always be first

bluewww:
damn its already fixed in master. But I guess I can still do the change you mentioned @accuminium

accuminium:
> damn its already fixed in master.

Yes, but one could argue that [the fix](https://github.com/pulp-platform/common_cells/commit/3364cbbd4f23b3ec56e51eb32fa7396f809b9d53) is wrong: `addr_decode` does not belong to Level 1 but to Level 0, since it depends on a file compiled in Level 0, `cf_math_pkg`.

bluewww:
I'll address this in another PR. I need this branch to be in a working state asap (due to ipaprox).

PR #106, opened 2020-12-01
src/id_queue.sv: Enables simultaneous input and output

flaviens:
When working on Fenrir we realized that the id_queue cannot simultaneously have stuff inserted and output. This commit fixes it, increasing the maximal achievable utilization from 50% to 100%. Tested with the provided testbench and in-system.

### DIFF
--- a/src/id_queue.sv
+++ b/src/id_queue.sv
@@ -14,11 +14,17 @@
 // queue preserves FIFO order.
 //
 // This ID queue implementation allows to either push (through the `inp_*` signals) or pop (through
-// the `oup_*` signals) one element per clock cycle. The `inp_` port has priority and grants a
-// request iff the queue is not full. The `oup_` port dequeues an element iff `oup_pop_i` is
-// asserted during an `oup_` handshake; otherwise, it performs a non-destructive read. `oup_data_o`
-// is valid iff `oup_data_valid_o` is asserted during an `oup_` handshake. If `oup_data_valid_o` is
-// not asserted, the queue did not contain an element with the provided ID.
+// the `oup_*` signals) one element per clock cycle (depending on the _FULL_BW_ operating mode
+// descibed below). The `inp_` port has priority and grants a request iff the queue is not full. The
+// `oup_` port dequeues an element iff `oup_pop_i` is asserted during an `oup_` handshake;
+// otherwise, it performs a non-destructive read. `oup_data_o` is valid iff `oup_data_valid_o` is
+// asserted during an `oup_` handshake. If `oup_data_valid_o` is not asserted, the queue did not
+// contain an element with the provided ID.
+//
+// The queue can work in two bandwidth modes:
+//  * !FULL_BW: Input and output cannot be performed simultaneously (max bandwidth: 50%).
+//  *  FULL_BW: Input and output can be performed simultaneously and a popped cell can be reused
+//    immediately in the same clock cycle. Area increase typically 5-10%.
 //
 // This ID queue additionally provides the `exists_` port, which searches for an element anywhere in
 // the queue. The comparison performed during the search can be masked: for every bit that is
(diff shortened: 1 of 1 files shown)


accuminium:
Thanks @flaviens, that is a very useful feature! Do you happen to know the timing and area impact? Thinking a step further: If the impact is non-negligible, I think this feature could be conditionally enabled with a parameter. Would this complicate the code a lot?

flaviens:
> Thanks @flaviens, that is a very useful feature! Do you happen to know the timing and area impact? Thinking a step further: If the impact is non-negligible, I think this feature could be conditionally enabled with a parameter. Would this complicate the code a lot?

@200MHz the area did not increase (what this commit does is mostly adding some control on already existing MUXes). I will do more careful checks, particularly at higher frequencies and will let you know!

thommythomaso:
Do you use GF22 for synthesis and did you choose a reasonable IO delay?

flaviens:
Hello!
I ran experiments on TSMC65 and GF22 technologies with 50% I/O delays. Between the two versions, there is an area increase of 3-8%. I have parametrized the module as recommended @accuminium, and made sure the default parameter leaves the module identical to the current (before this version).

zarubaf:
Hey, @flaviens thanks for the PR. I have a couple of comments:

- Can we test the new configuration somehow? i.e., adapting the existing testbench to also test the new parameter configuration.
- Is the standard configuration (`id_queue_tb`) working for you? I seem to get many errors, so just wondering whether this is a set-up issue from my side or if there are really some problems lingering.
- Is there any way we can make the change a little less intrusive and more additive to the original function? Just asking whether we can make the PR a little easier reviewable.
- Can you please re-base on the latest master. @WRoenninger has removed the `x` assignments there. Thanks.

accuminium:
>     * Can you please re-base on the latest master. @WRoenninger has removed the `x` assignments there. Thanks.

I had promised to do that, since I wanted the @WRoenninger's fix urgently.  You can find the [rebased commit](https://github.com/pulp-platform/common_cells/commit/2c094d546372ad31a5a5d77586516fe4d49edb57) on the [`flaviens-master` branch in this repository](https://github.com/pulp-platform/common_cells/commits/flaviens-master) (since I had no permission pushing to your repository).

accuminium:
>     * Is the standard configuration (`id_queue_tb`) working for you? I seem to get many errors, so just wondering whether this is a set-up issue from my side or if there are really some problems lingering.

I can confirm that `id_queue_tb` throws many errors on this branch but not on `master`.

flaviens:
Hi!

> Hey, @flaviens thanks for the PR. I have a couple of comments:
> 
>     * Can we test the new configuration somehow? i.e., adapting the existing testbench to also test the new parameter configuration.

Yes, through the existing testbench. We may want to run the testbench for the two possible values of `!FULL_BW`.

> 
>     * Is the standard configuration (`id_queue_tb`) working for you? I seem to get many errors, so just wondering whether this is a set-up issue from my side or if there are really some problems lingering.

There were indeed some rare errors, thank you for pointing this out! This is now solved.
> 
>     * Is there any way we can make the change a little less intrusive and more additive to the original function? Just asking whether we can make the PR a little easier reviewable.

This was done very incrementally on what actually exists. Referring to the previous discussion, I have added a parameter to get the exact same delays as before this commit. This justifies the large `if (!FULL_BW) begin` starting on line 204, that is almost untouched compared to the previous version. The `else` part has the same logic but has to additionally manage some corner cases, such as when data comes in and data comes out simultaneously, and there is no free space except the space currently freed by popping.
If you take some time understanding again the original version first, you may find out that it is in fact quite incremental.

I believe that the `if (!FULL_BW) begin` block (as opposed to its `else` block) should be removed once it is ensured that all the current designs support this commit. This will reduce the source code size compared to what's in this commit for the moment.

> 
>     * Can you please re-base on the latest master. @WRoenninger has removed the `x` assignments there. Thanks.

Done.

> 
> 
> And sorry it took so long for me to get back to you ;)

No worries, I'm available if you still have comments!

zarubaf:
Sorry that this slipped through the cracks for sooo long...

PR #108, opened 2020-12-31
Replace NDEBUG define with an XSIM guard

olofk:
Replaced the previously added NDEBUG define with a tool-specific XSIM define analogous to how the VERILATOR define is used


zarubaf (approved):
Awesome. Thanks!

PR #111, opened 2021-01-28
Bump common_verification and tech_cells_generic dependency

micprog:
Updates the following dependencies:
- `tech_cells_generic` (v0.1.1 -> v0.2.3)
- `common_verification` (v0.1.1 -> v0.2.0)
    - requires renaming of some parameters

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -10,8 +10,8 @@ package:
     - "Wolfgang Roenninger <wroennin@student.ethz.ch>"
 
 dependencies:
-  common_verification: { git: "https://github.com/pulp-platform/common_verification.git", version: 0.1.1 }
-  tech_cells_generic: { git: "https://github.com/pulp-platform/tech_cells_generic.git", version: 0.1.1 }
+  common_verification: { git: "https://github.com/pulp-platform/common_verification.git", version: 0.2.0 }
+  tech_cells_generic: { git: "https://github.com/pulp-platform/tech_cells_generic.git", version: 0.2.3 }
 
 export_include_dirs:
   - include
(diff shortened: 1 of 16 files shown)


zarubaf:
Thanks, @micprog. Would you mind checking the test benches by running `simulate.sh` in the `test` folder?

I get a couple of errors that seem to be related to the dependency upgrade.

```
# ** Error (suppressible): /home/zarubaf/eth/common_cells/test/fifo_tb.sv(179): (vopt-2732) Module parameter 'ClkPeriod' not found for override.
```
```
# ** Error (suppressible): /home/zarubaf/eth/common_cells/test/id_queue_tb.sv(66): (vopt-2732) Module parameter 'ClkPeriod' not found for override.
```
```
# ** Error (suppressible): /home/zarubaf/eth/common_cells/test/stream_to_mem_tb.sv(129): (vopt-2732) Module parameter 'ClkPeriod' not found for override.
```

micprog:
Everything seems to work for me. Due to the update in dependencies, you may need to run 
```bash
bender update
```
before executing the test script, as this will re-resolve the dependencies and then fetch the correct version. If you want, I can add this command to the `simulate.sh` script, then it would always work when updating these dependency references. The error you're encountering is likely due to `common_verification` not yet updated to v0.2.0, as one of the differences is renaming of these parameters.

zarubaf:
Thanks, that indeed was the problem. I've also removed the `timeunit` arguments as they always make problems and can be set via command-line arguments.

micprog:
Thanks! I also noticed the issue with `timeunit`, an override could be added to the `bender-script` command, similar to [`tech_cells_generic`](https://github.com/pulp-platform/tech_cells_generic/blob/master/scripts/compile_vsim.sh) if needed, but this has some adverse effects (e.g. `cdc_2phase_tb`).

A new release would be easiest to manage the added change, thank you!

PR #113, opened 2021-02-17
rr_arb_tree: update assertion to allow flushing locked decision

niwis:
When `rr_arb_tree` is locked, allow flushing the decision (i.e. do not throw an error). Otherwise, asserting `flush_i` with `LockIn` set will fail.


zarubaf (approved):
LGTM!

PR #114, opened 2021-03-03
Verilator issue with sub_per_hash

suehtamacv:
The latest Verilator version has issues with some constructs used in the `sub_per_hash` module: 

``` 
%Error: common_cells/src/sub_per_hash.sv:54:42: Expecting expression to be constant, but can't determine constant for FUNCREF 'get_permutations'
                                                                                                      : ... In instance common_cells/src/sub_per_hash.sv:109:23: ... Location of non-constant ASSIGN: Array select LHS isn't simple variable
        common_cells/src/sub_per_hash.sv:54:42: ... Called from get_permutations() with parameters:
           seed = 32'h47df4
```

The issue is that `XorStages` and `PERMUTATIONS` are declared as `localparam`'s, whose value depends on the result of a function. The issue is solved by declaring those two variables as signals.

This fix is between `ifdef VERILATOR` guards, but there is no issue with using this construct with other EDA tools.

### DIFF
--- a/src/sub_per_hash.sv
+++ b/src/sub_per_hash.sv
@@ -51,12 +51,14 @@ module sub_per_hash #(
 
   // typedefs and respective localparams
   typedef int unsigned perm_lists_t [NoRounds][InpWidth];
-  localparam perm_lists_t PERMUTATIONS = get_permutations(PermuteKey);
+  perm_lists_t Permutations;
+  assign Permutations = get_permutations(PermuteKey);
   // encoding for inner most array:
   // position 0 indicates the number of inputs, 2 or 3
   // the other positions 1 - 3 indicate the inputs
   typedef int unsigned xor_stages_t [NoRounds][InpWidth][3];
-  localparam xor_stages_t XorStages = get_xor_stages(XorKey);
+  xor_stages_t XorStages;
+  assign XorStages = get_xor_stages(XorKey);
 
   // stage signals
   logic [NoRounds-1:0][InpWidth-1:0] permuted, xored;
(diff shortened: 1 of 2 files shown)


zarubaf:
Then I'd suggest we remove the ifdef and make this the universal approach, would that work for you? Constant folding should figure out that there is an opportunity to simplify this :-)

suehtamacv:
I push-forced, removing the `ifdef VERILATOR` guard.

src/sub_per_hash.sv
```diff
@@ -51,12 +51,14 @@ module sub_per_hash #(
 
   // typedefs and respective localparams
   typedef int unsigned perm_lists_t [NoRounds][InpWidth];
-  localparam perm_lists_t PERMUTATIONS = get_permutations(PermuteKey);
+  perm_lists_t PERMUTATIONS;
```
zarubaf, line 54: There is only one small nitpick (sorry about that). Now that this is a signal it should be `snake_case` I guess. Same for the `XorStages` below.

WRoenninger:
When I remember correctly I have seen something similar actually when implementing it.

Just thinking, is it an issue that the definition of the function is below assigning to the localparam? Would it help to move the function declarations between the `typedef` and the localparam assign? 

I would like to keep it as a localparam if possible for showing that the permutation is not changing. Am not sure why some tools are not recognizing it as a constant function.

niwis:
@suehtamacv is this PR still active or should it be closed?

suehtamacv:
> @suehtamacv is this PR still active or should it be closed?

It is still active. Just did the changes required.

niwis:
Thank you a lot! I think the signal name also needs to be updated below: https://github.com/pulp-platform/common_cells/blob/5d112fc197c259f20d0aca5a575b5be08af23828/src/sub_per_hash.sv#L69-L74

Finally, could you also add a note in the `CHANGELOG`?

PR #115, opened 2021-04-14
include/common_cells/registers.svh: Add NO_SYNOPSYS_FF macro

flaviens:
Hi!

Some open source tools for synthesis, for instance [sv2v](https://github.com/zachjs/sv2v) chained with [Yosys](https://github.com/YosysHQ/yosys/) do not understand the `/``* synopsys ... *``/` syntax. Since they are not Verilator, another macro could be very helpful.
Defining the `VERILATOR` is generally not a good solution because this macro also enforces simulation code instead of synthesis code.

Maybe there are more elegant solutions, feel free to propose!

### DIFF
--- a/include/common_cells/registers.svh
+++ b/include/common_cells/registers.svh
@@ -83,7 +83,9 @@
 // __reset_clk: reset input
 `define FFSR(__q, __d, __reset_value, __clk, __reset_clk) \
   `ifndef VERILATOR                       \
+  `ifndef NO_SYNOPSYS_FF                  \
   /``* synopsys sync_set_reset `"__reset_clk`" *``/       \
+    `endif                        \
     `endif                        \
   always_ff @(posedge (__clk)) begin                      \
     __q <= (__reset_clk) ? (__reset_value) : (__d);       \
@@ -97,7 +99,9 @@
 // __reset_n_clk: reset input
 `define FFSRN(__q, __d, __reset_value, __clk, __reset_n_clk) \
     `ifndef VERILATOR                       \
+    `ifndef NO_SYNOPSYS_FF                  \
   /``* synopsys sync_set_reset `"__reset_n_clk`" *``/        \
+    `endif                        \
     `endif                        \
   always_ff @(posedge (__clk)) begin                         \
     __q <= (!__reset_n_clk) ? (__reset_value) : (__d);       \
@@ -170,7 +174,9 @@
 // __reset_clk: reset input
 `define FFLSR(__q, __d, __load, __reset_value, __clk, __reset_clk)       \
     `ifndef VERILATOR                       \
+    `ifndef NO_SYNOPSYS_FF                  \
   /``* synopsys sync_set_reset `"__reset_clk`" *``/                      \
+    `endif                        \
     `endif                        \
   always_ff @(posedge (__clk)) begin                                     \
     __q <= (__reset_clk) ? (__reset_value) : ((__load) ? (__d) : (__q)); \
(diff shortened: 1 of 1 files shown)


olofk:
Yes! Very much agree. I brought up a similar issue a while ago (https://github.com/pulp-platform/axi/pull/131#issuecomment-753227543) when adding support for other tools. Make the defines reflect the purpose (e.g. `DISABLE_FORMAL`) and let the build system handle setting the correct defines depending on the environment

zarubaf:
Thanks for resurrecting @olofk - that slipped through the cracks. So if we follow along the lines as in https://github.com/pulp-platform/axi/pull/131#issuecomment-753227543 that would mean we guard the Synopsys macros with an `ifdef SYNOPSYS`?

olofk:
I think two things here.
1. Generally, I think it would be better to have something like
```
`ifdef VERILATOR
  AVOID_TRICKY_CODE
`endif

`ifdef AVOID_TRICKY_CODE
...
`endif
```
for all the PULP code because that will be backwards-compatible with the current code but still allow setting a defne externally to enable/disable some parts. Something like https://github.com/pulp-platform/common_cells/pull/107 (Note: For some reason that I can't remember anymore, I kind of reverted this with https://github.com/pulp-platform/common_cells/pull/108 . Maybe it was because we wanted to stay close to the OpenTitan code?)

2. I took another look at this particular issue since I was a bit suprised that the macros is causing issues. All tools I'm aware of would just ignore the comments. I think the actual problem is that sv2v has problems parsing ```  /``* any_random_attribute `"__clear`" *``/                       \``` So again, I don't think it should be guarded with a SYNOPSYS ifdef since other tools might very well pick up the attribute correctly. But instead add a `AVOID_ATTRIBUTES_IN_MACRO_BUT_PERHAPS_CALL_IT_SOMETHING_SHORTER`  define

olofk:
Ok, so I'm diving deep into this rabbit hole now and thought we might get rid of the issue altogether if we used proper attributes instead of those nasty meta-comments that were popular during the previous millenium. I can't find any definite proof of what the syntax should look like though but would it be possible to see if this change produces the equivalent code for the tools you care about? It makes sv2v happier at least
```
 // __clk: clock input
 // __arst_n: asynchronous reset
 `define FFLARNC(__q, __d, __load, __clear, __reset_value, __clk, __arst_n) \
-    `ifndef VERILATOR                       \
-  /``* synopsys sync_set_reset `"__clear`" *``/                       \
-    `endif                        \
-  always_ff @(posedge (__clk) or negedge (__arst_n)) begin                 \
+ (* sync_set_reset="__clear" *) always_ff @(posedge (__clk) or negedge (__arst_n)) begin                 \
     if (!__arst_n) begin                                                   \
       __q <= (__reset_value);                                              \
     end else begin                                                         \
```

zarubaf:
Okay, rabbit hole alarm! @stmach did the previous rabbit hole exploration - maybe he still has input on this? Should be fairly straightforward to try this on our end.

zarubaf:
A first test showed that it isn't picked up by Synopsys by default. I remember that there is some switch that makes it pick up the RTL attributes. I'll dig into the user manual.

zarubaf:
No luck, the `sync_set_reset` primitive can only be used with the `synopsys` qualifier - anything else isn't picked up :/

olofk:
My love for EDA tools grow stronger every day. Ok, to move this forward I would suggest then to do a variation of @flaviens patch but to _replace_ the `ifndfef VERILATOR` with `ifndef NO_SYNOPSYS_FF` and then in the beginning of the file add

```
`ifdef VERILATOR
`define NO_SYNOPSYS_FF 1
`endif
```

This way we can set `NO_SYNOPSYS_FF` when parsing e.g. with sv2v and don't need extra guards for different tools. Or maybe we'll just preproess with Morty, which I assume will handle this just fine :)

stmach:
Hi there! AFAIR I've applied the meta-comment syntax to match what was expected by Synopsys' documentation at the time, and haven't really checked many other tools beyond Verilator.

I think @olofk's suggestion makes sense to keep things somehow tidy for now.

zarubaf:
Guys (@olofk, @stmach, @flaviens) can you check whether #117 does the job for you?

olofk:
I'll check if sv2v is happy with #117. Also pinging @zeeshanrafiq23, who had the sv2v problems in his design originally, for extra tests

olofk:
ok, assuming you meant #118, not #117 sv2v is happy when setting `-D NO_SYNOPSYS_FF=1` so all good from that perspective. Don't know if the code actually runs but I think that's really a different problem

zarubaf:
Should be fixed now, thanks!

PR #117, opened 2021-05-20
registers.svh: Merge explicit and implicit macros

stmach:
Merged explicit and implicit macros into `` `FF `` and `` `FFL `` for
asynchronous FFs without and with load-enable, respectively. The old
macros `` `FFARN `` and `` `FFLARN `` are retained for compatibility but
a deprecation notice was added to them.


zarubaf (approved):
Awesome, LGTM!

PR #120, opened 2021-06-13
Fix port list copy paste bug

meggiman:
The introduction of a new port to an existing module is not backward compatible and
thus breaks every other repository that uses the spill_registers from common
cells. I guess the intention was to not introduce this port and have the
spill_register module act as just a wrapper.

After merging this commit, we should immediately release a new minor version to restore compatibility.


zarubaf:
Indeed, and I thought I've fixed this - must be a rebase mishap. Thanks for addressing this.

PR #124, opened 2021-09-05
Release v1.23.0

SamuelRiedel:
Create a new release with the latest changes.

Unfortunately, I don't have permission to push directly to master. So either we ff-merge this, or @zarubaf, you could cherry-pick this commit.


zarubaf (approved):
👍

PR #127, opened 2022-01-18
add round-robin distributor

micprog:


src/rr_distributor.sv
```diff
@@ -0,0 +1,79 @@
...
+module rr_distributor # (
+    /// Number of outputs to distribute to.
+    parameter int unsigned NumOut    = 1,
+    /// Data width of the payload in bits. Not needed if `payload_t` is overwritten.
+    parameter int unsigned Width     = 1,
+    /// Data type of the payload, can be overwritten with a custom type. Only use of `Width`.
+    parameter type         payload_t = logic [Width-1:0],
+    /// Dependent parameter, do **not** overwrite.
+    /// Width of the selected index
+    parameter int unsigned IdxWidth  = (NumOut > 32'd1) ? unsigned'($clog2(NumOut)) : 32'd1,
```
accuminium, line 25:
```suggestion
    parameter int unsigned IdxWidth  = cf_math_pkg::idx_width(NumOut),
```

src/rr_distributor.sv
```diff
@@ -0,0 +1,79 @@
...
+// Description: round robin distributor
+
+/// The rr_distributor forwards requests to individual outputs in a round robin fashion.
+module rr_distributor # (
+    /// Number of outputs to distribute to.
+    parameter int unsigned NumOut    = 1,
+    /// Data width of the payload in bits. Not needed if `payload_t` is overwritten.
+    parameter int unsigned Width     = 1,
+    /// Data type of the payload, can be overwritten with a custom type. Only use of `Width`.
+    parameter type         payload_t = logic [Width-1:0],
```
accuminium, line 22:
```suggestion
    parameter type         data_t = logic [Width-1:0],
```
In most modules in this repository, we simply name the *data type of the payload* `data_t`, since there is no other data type at the interface that it could be confused with

Bender.yml
```diff
@@ -43,6 +44,7 @@ sources:
   - src/plru_tree.sv
   - src/popcount.sv
   - src/rr_arb_tree.sv
+  - src/rr_distributor.sv
```
accuminium, line 47: Do not forget to add it to `src_files.yml`

src/rr_distributor.sv
```diff
@@ -0,0 +1,79 @@
...
+        assign valid_o[0] = valid_i;
+        assign ready_o    = ready_i[0];
+        assign sel_o      = 1'b0;
+    end else begin : gen_arb
+
+        idx_t rr_d, rr_q;
+        logic one_ready;
+
+        assign rr_d = (valid_i & one_ready) ?
+            ((rr_q == idx_t'(NumOut-1)) ? '0 : rr_q + 1'b1) : rr_q;
```
accuminium, line 53: I strongly suggest to write this as `always_comb` instead of as nested ternary statements, and add comments describing why assignments are made

src/rr_distributor.sv
```diff
@@ -0,0 +1,79 @@
...
+
+        assign one_ready = |ready_i;
+
+        always_comb begin : proc_arbitration
+            valid_o = '0;
+            ready_o = 1'b0;
+            if (ready_i[rr_q]) begin
+                valid_o[rr_q] = valid_i;
```
accuminium, line 61: This does not guarantee that a `valid_o` remains high once asserted until the handshake, correct? Many of our modules that work with valid/ready assume this property (borrowed from AXI handshakes), because it makes handling inputs much easier

andreaskurth (changes requested):

src/rr_distributor.sv
micprog: Doesn't it? `valid_o` will only be high if `ready_i` is high (and `valid_i` is high), meaning that whenever `valid_o` is high the handshake completes, and isn't high otherwise.

accuminium: You are totally right, I don't remember why I asked this. :slightly_smiling_face: 

But I see another problem with this code: For the output handshake, `valid_o` depends on `ready_i`, but the rule in valid/ready handshaking is that `ready` can depend on `valid`, but not vice-versa (to prevent combinatorial loops).

src/rr_distributor.sv
```diff
@@ -0,0 +1,87 @@
...
+        assign ready_o    = ready_i[0];
+        assign sel_o      = 1'b0;
+    end else begin : gen_arb
+
+        idx_t rr_d, rr_q;
+        logic one_ready;
+
+        always_comb begin : rr_next
+            rr_d = rr_q;
+            if (valid_i && one_ready) begin
```
accuminium, line 54: Is it intentional that the counter increases in cycles when there is no handshake? Or should this condition be `valid_i && ready_i[rr_q]` (or, equivalently after the change below `valid_i && ready_o`)?

micprog: @thommythomaso is this intentional?
It being intentional would make sense to me, as in this case a valid will be served within a few cycles if any subordinate is available, even if the next in the rr distribution scheme is not ready yet (e.g. stalled for 100s of cycles). With the change discussed above, this may cause deassertion of the valid signal however, which would be against the ready/valid rules. However, not implementing the change above is also against these rules. I currently don't see a clean way to implement this feature while conforming to ready/vaild handshake rules.

accuminium: Indeed. I suggest we clarify what "forwards requests to individual outputs in a round robin fashion" means in conjunction with `valid`/`ready` handshaking.

If that is not possible or does not make sense, an option could be to relax to `req`/`gnt` handshaking (where `req` can be deasserted without `gnt`, but `req` must still not depend on `gnt`).

src/rr_distributor.sv
```diff
@@ -0,0 +1,93 @@
...
+    parameter type         data_t    = logic [Width-1:0],
+    /// Setting StrictRR enforces a strict round-robin distribution
+    /// If set to 1'b1, the rr_distributor will distribute the requests to the next output
+    ///   in line, irrespective if this output is ready or not, irrespective if another
+    ///   output could accept the request. The transaction will wait until the next one in
+    ///   line accepts the handshake.
+    /// If set to 1'b0, the rr_distributor will step through the outputs if one is ready
+    ///   but the current one is not. This can reduce wait time for the input.
+    ///   **THIS IS NOT COMPLIANT AS IT MAY DE-ASSERT VALID WITHOUT A PROPER HANDSHAKE**
+    parameter bit          StrictRR  = 1'b0,
```
accuminium, line 31: Thanks for the clarification!

Now, is `StrictRR = 1'b0` an actual use case? Also asking @thommythomaso?

micprog: The behavior outlined by `StrictRR = 1'b0` is currently used in the HERO dma implementation as far as I know, allowing for automatic distribution of dma transfers of different lengths. This way, if many requests are issued (> number of channels) and some are extremely long, the next transfer will be issued as soon as a dma is available, instead of waiting for the next channel in line to be available.

accuminium: I understand, thanks for the explanation.

Wouldn't it make sense to *reverse* the handshake on the output interface? That is

```sv
// output stream
input  logic  [NumOut-1:0] req_data_valid_i,
output logic  [NumOut-1:0] req_data_ready_o,
output data_t [NumOut-1:0] data_o,
output idx_t               sel_o
```

Downstream modules could then request data through `req_data_valid_i` as soon as they are available, and the `rr_distributor` can select one of them and set its `data_req_ready_o`.

phsauter:
I am closing this for being inactive for a very long time. If you think there is still merit here feel free to open it again.

PR #128, opened 2022-01-27
sync: Extend with data type parameter

luca-valente:
The `sync` module works for just one bit. `sync_t` can synchronize
more bits with one instance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,9 @@ The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## Unreleased
+### Added
+- `sync`: Add data type parameter, so this module can be used on any signal instead of just
+  single-bit `logic`s.
 
 ## 1.23.0 - 2021-09-05
 ### Added
--- a/src/sync.sv
+++ b/src/sync.sv
@@ -11,16 +11,17 @@
 // Antonio Pullini <pullinia@iis.ee.ethz.ch>
 
 module sync #(
+    parameter type T = logic,
     parameter int unsigned STAGES = 2,
-    parameter bit ResetValue = 1'b0
+    parameter T ResetValue = '0
 ) (
     input  logic clk_i,
     input  logic rst_ni,
-    input  logic serial_i,
-    output logic serial_o
+    input  T     serial_i,
+    output T     serial_o
 );
 
-   logic [STAGES-1:0] reg_q;
+    T [STAGES-1:0] reg_q;
 
     always_ff @(posedge clk_i, negedge rst_ni) begin
         if (!rst_ni) begin


andreaskurth (commented):
Thanks for this suggestion, @luca-valente!

Why would you like to add a new module, instead of adding a `T` parameter to the `sync` module and setting its default value to `logic` for backward compatibility?

luca-valente:
No real reason actually :smile: 
Having just the `sync` module, as you suggested, just makes more sense!

luca-valente:
@accuminium pushed the needed modifications!

accuminium:
> @accuminium pushed the needed modifications!

Thanks! :+1: I took the liberty to squash the two commits into one, fix the whitespace lint errors, and write a more detailed commit message.

accuminium:
I agree with this implementation, but it's up to @niwis to decide.

By the way, one must be very careful with synchronizing multiple bits this way, see, e.g., *Data Consistency in Vectored Acquisition* in the *Top-Down Digital VLSI Design* book. If you know exactly what you are doing and are able to enforce the constraints that make this work, this multi-bit synchronizer can be useful. But it's also very easy to get it wrong and get inconsistent data. @meggiman: What do you think about this extension?

meggiman:
Indeed @accuminium,  synchronizing multiple bits is something you rarely ever do using multiple synchronizers. In fact I don't see much use for this module except for the synchronization of gray coded values (e.g. CDC FIFO write/read pointers). Whenever you need to synchronize data across a CDC where more than a single bit can change at a time, either some sort of handshaking or dual-clock FIFO/memory must be used. Except for the write/read pointers in a dual-clock FIFO, there is no need to synchronize the actual data in those cases since your handshaking protocol ensures data stability during the sampling window. Nonetheless, I can certainly see that this module could help to clean up the existing CDC Fifo code a bit (we could replace the generic for loop to instantiate the individual pointer bits with a single instance). @luca-valente Do you have a specific use case in mind for the multi-bit synchronizer?

luca-valente:
Hi @meggiman,
yes, I actually have a use case. I am building a stand-alone macro of the hyperbus module. To do so, we want to expose asynchronous interfaces, as it's done for the cluster, for example. Indeed, for the `AXI SLAVE PORT` I added the [axi_cdc_dst](https://github.com/pulp-platform/axi/blob/master/src/axi_cdc_dst.sv) before it.
On the other hand, the udma interface is divided in 5 channels:  
* APB port for config registers
* tx channel
* [tx channel control](https://github.com/pulp-platform/udma_core/blob/master/rtl/core/udma_tx_channels.sv#L227)
* rx channel
* [rx channel control](https://github.com/pulp-platform/udma_core/blob/master/rtl/core/udma_rx_channels.sv#L306)

For the APB and tx/rx channel it is straightforward to utilize the available `cdc_gray_fifo`, while for the channel controls there is no `ready/valid` handshake to exploit. That's why I need to synchronize the transfer meta-data with `sync` module, and I wanted to avoid the same `for` loop you mentioned.
What do you think?

meggiman:
That sounds very dangerous. Maybe we should discuss that in a dedicated meeting and resume this thread afterwards.

meggiman:
After some discussion with @luca-valente we came to the conclusion that this modification is not necessary for his purposes. While a parametric width sync cell would simplify some of our CDC FIFO's it bears the risk of future misuse in unsafe synchronization schemes since synchronizing more than one bit of correlated, non-gray coded data through a sync cell is almost always a design flaw. After checking with @accuminium I'm thus closing this PR. @niwis feel free to reopen if you disagree.

PR #129, opened 2022-02-10
iso-crossings: Fix reset polarity in assertions

paulsc96:
The `disable iff` conditions in these assertions did not account for the active-low reset polarity.


accuminium (approved):
The implementation LGTM, thanks @paulsc96! I suggest you add a Changelog entry, though.

It's up to @niwis as maintainer to decide on the merge, of course.

PR #130, opened 2022-02-21
Edge propagator ack

glaserf:
Added a version of the edge propagator that has an acknowledge port which flags if the edge has actually propagated to the receiving side. Useful if the initiating circuit can only safely continue if the described condition is met.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -136,5 +136,6 @@ sources:
   - src/deprecated/fifo_v1.sv
 
   # Depend on deprecated modules
+  - src/edge_propagator_ack.sv
   - src/edge_propagator.sv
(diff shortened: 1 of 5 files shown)


niwis:
Thank you @glaserf! Could you also add the module to `README` and `CHANGELOG`?
One general question: There seems to be a lot of common code with `edge_propagator`. Do you think it would make sense to update `edge_propagator` such that it wraps `edge_propagator_ack` (leaving `ack_o` unconnected) to avoid code duplication?

glaserf:
Cool! I implemented both suggestions.

niwis (approved):
LGTM, thanks a lot @glaserf and @accuminium !

PR #131, opened 2022-03-17
Add optional seed parameter to stream_delay module

meggiman:
The stream_delay module optionally contains an LFSR to randomize the delays. If we instantiate more than one of these modules as verification IP, it would be useful being able to set the seed of each LFSR individually. This PR exposes the LFSR seed parameter to the toplevel and assigns it the default value to maintain backward compatibility.


niwis (approved):
Thank you a lot, looks good to me. Could you also add an entry in the `CHANGELOG`?

PR #132, opened 2022-03-21
Configurable arbitrary integer clock divider

meggiman:
The existing `clk_div` clock divider module has some flaws:

- The module is not configurable. I.e. the division factor can only be chosen at elaboration time. However, in many of our IPs we require an at-runtime configurable clock divider.
- The generated output clock has an extremely unbalanced duty cycle. While this might be ok for strictly positive-edge triggered logic, mistakenly using this clock divider for logic that makes use of both clock edges (e.g. latch based standard cell memories, DDR links etc.) would result in serious timing issues

This PR adds a new configurable clock divider IP (`clk_int_div`) that is guaranteed to always generate a 50% duty cycle output clock regardless divider value (even and odd division is both supported). To that end it uses two toggle flip-flops with 90° phase shift to generate the odd clock division with 50% duty cycle. Furthermore, the module is at runtime configurable with parametric max division factor. The configuration interface is (valid-ready) handshaked and automatically gates the clock during configuration changes to prevent any clock glitching at the output. The design was verified using a new dedicated testbench (`clk_int_div_tb`). 

All sequential logic is strictly clocked by the input clock only and the fan-in of the output clock  exclusively consists of sequential elements (t_flip_flops, latched control signals) and dedicated clock gating cells from `tech_cells_generic` (verified by trial synthesis). The only exception to this rule is the combinational `test_mode_en_i` that enables clock bypassing for DFT. 

Besides adding the new `clk_int_div` this PR deprecates the existing `clk_div` module. Existing designs using `clk_div` will observe a warning message (can be disabled with optional design parameter) during elaboration explaining the flaws of the now deprecated implementation.


meggiman:
This PR would replace PR #46 solving the mentioned issues with the previous implementation

niwis (approved):
Great contribution, thank you a lot @meggiman!

PR #133, opened 2022-03-22
lzc: Fix compatibility issue with Verilator

SamuelRiedel:
The previous version led to problems with Verilator for some parameterizations. While this change does not impact the module's behavior, it is cleaner and makes Verilator work.

There is no behavioral change, as the modified line is only executed if `WIDTH > 1`, in which case `$clog2(WIDTH) = idx_width(WIDTH)`


niwis:
@SamuelRiedel is this PR ready? If so, could you open it so that I can squeeze it into the next release?

SamuelRiedel:
Yes, this is ready. Thanks.

PR #136, opened 2022-04-12
Small typofix for clk div in src_files.yml

glaserf:
Nonexisting file referenced


niwis (approved):
Thanks a lot @glaserf!

PR #137, opened 2022-04-13
Align src_files.yml with Bender

glaserf:
Some newly added files, mostly related to CDCs, seem to have not made it into src_files.yml


niwis (approved):
Thank you again @glaserf!

PR #139, opened 2022-05-03
add support for fusesoc and vivado xsim

skokvermon:


niwis:
`common_cells.core` was updated in #141 and #143. Regarding the xsim typedef: could you elaborate on this change in a separate PR? afaik `XSIM` should be the correct define. Finally, if this needs to be changed, this should be done globally in all source files where it is used.

skokvermon:
Sorry for late reply. I confirm XSIM is not defined with vivado xsim simulator
macro to use is XILINX_SIMULATOR
This change has to be made in all pulp code...

official source is in UG900 page 
https://docs.xilinx.com/r/2021.2-English/ug900-vivado-logic-simulation/Predefined-Macros

PR #140, opened 2022-05-06
stream_register: Replace FIFO with FFs

stmach:
The use of `fifo_full` to determine readiness caused a stall on every second clock cycle. This commit replaces the FIFO implementation with a simple register with proper propagation of the `ready` signal.
Furthermore, bubbles in the stream are resolved combinatorially now.

### DIFF
--- a/src/stream_register.sv
+++ b/src/stream_register.sv
@@ -29,7 +29,7 @@ module stream_register #(
     input  logic    ready_i,
     output T        data_o
 );
-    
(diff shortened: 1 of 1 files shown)


stmach:
Oops, messed up the commit message, should have been `stream_register` 😉

niwis (approved):
Thank you a lot @stmach, this is a much-needed fix!

PR #143, opened 2022-08-08
Fix FuseSoc core version

hossein1387:
It seems `common_cells.core` is out of date for some time now. However there is at least[ one other IP](https://github.com/pulp-platform/axi/blob/db457aa8d8996d87ca72486a9b15873042e04d38/axi.core#L54) in PULP that expects newer version of `common_cells` through FuseSoc. This PR should fix that.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## 1.24.1 - 2022-04-13
 ### Fixed
+- Fix FuseSoc core version in `common_cells.core`
 - Fix typos in `Bender.yml` and `src_files.yml`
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 
 ## 1.24.1 - 2022-04-13
 ### Fixed
+- Fix FuseSoc core version in `common_cells.core`
```
niwis, line 9: This version is already released, so the changelog is frozen
```suggestion
```

niwis (changes requested):
Hi @hossein1387, thank you for fixing this. 1.24.1 is already released, so I think we should not modify its changelog. I will keep an eye on fusesoc versions in future releases.

niwis (approved):

PR #144, opened 2022-08-08
Add `addr_decode_napot`

paulsc96:
Add `addr_decode_napot`, which uses a base address and mask instead of a start and end address.

TODO:

* [x] Avoid `addr_decode_napot` copypasta by properly extending and wrapping `addr_decode`
* [ ] Check refactor in simulation

### DIFF
--- a/src/addr_decode_napot.sv
+++ b/src/addr_decode_napot.sv
@@ -1,0 +1,119 @@
+// Copyright 2019 ETH Zurich and University of Bologna.
+// Copyright and related rights are licensed under the Solderpad Hardware
+// License, Version 0.51 (the "License"); you may not use this file except in
+// compliance with the License.  You may obtain a copy of the License at
+// http://solderpad.org/licenses/SHL-0.51. Unless required by applicable law
+// or agreed to in writing, software, hardware and materials distributed under
+// this License is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR
+// CONDITIONS OF ANY KIND, either express or implied. See the License for the
+// specific language governing permissions and limitations under the License.
+
+// Author: Wolfgang Roenninger <wroennin@ethz.ch>
+
+/// Address Decoder: Maps the input address combinatorially to an index.
+/// The address map `addr_map_i` is a packed array of rule_t structs.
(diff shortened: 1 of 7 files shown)


src/addr_decode_napot.sv
```diff
@@ -0,0 +1,93 @@
...
+  ///  - `mask`:  set for bits which are to be checked to determine a match.
+  parameter type         rule_t    = logic,
+  /// Dependent parameter, do **not** overwite!
+  ///
+  /// Width of the `idx_o` output port.
+  parameter int unsigned IdxWidth  = cf_math_pkg::idx_width(NoIndices),
+  /// Dependent parameter, do **not** overwite!
+  ///
+  /// Type of the `idx_o` output port.
+  parameter type         idx_t     = logic [IdxWidth-1:0]
```
niwis, line 44: The dependent parameters be removed
```suggestion
  parameter type         rule_t    = logic
```

src/addr_decode_napot.sv
```diff
@@ -0,0 +1,93 @@
...
+  } rule_range_t;
+
+  addr_decode #(
+    .NoIndices ( NoIndices    ) ,
+    .NoRules   ( NoRules      ),
+    .addr_t    ( addr_t       ),
+    .rule_t    ( rule_range_t ),
+    .Napot     ( 1            ),
+    .IdxWidth  ( IdxWidth     ),
+    .idx_t     ( idx_t        )
```
niwis, line 82: Same here
```suggestion
    .Napot     ( 1            )
```

niwis (changes requested):
Thanks a lot! Could you also add the files to `src_files.yml` and `common_cells.core`?

src/addr_decode_napot.sv
paulsc96: Actually, this turns out to be necessary after all, since some of the IO is `idx_t` and needs the dependent parameters. But it was good to double-check ;)

paulsc96:
> Thanks a lot! Could you also add the files to `src_files.yml` and `common_cells.core`?

done :)

niwis (approved):

PR #146, opened 2022-08-09
Remove `program` blocks from testbenches for VCS compatibility

paulsc96:
The `program` blocks in some testbenches cause elaboration issues in VCS. Since we do not rely on the scheduling guarantees they provide, they are functionally redundant, and thus we can remove them.


niwis (approved):
Thanks a lot!

PR #147, opened 2022-08-09
Use glitch-free `tc_clk_mux` muxes in `rstgen_bypass`

paulsc96:
The current `rstgen_bypass` uses behaviorally-inferred multiplexing logic to switch between test mode and regular operation. Although `test_mode_i` is usually quasistatic, the generated reset may exhibit glitches if this does not hold for any reason or if the inferred multiplexing logic produces glitches when the reset itself toggles.


niwis (approved):
LGTM!

PR #148, opened 2022-08-09
Exclude `cb_filter` from Vivado IP packager projects

paulsc96:
Remove `cb_filter` and `cb_filter_pkg` from from Vivado IP packager project sources due to compatibility issues. Note that this does *not* remove it from sources when using Vivado by itself.


niwis (approved):
Thank you! Tracking IPX compatibility of `cb_filter` in https://github.com/pulp-platform/common_cells/issues/149

Issue #150, opened 2022-08-11
Stream join dynamic IP with independent input handshakes


colluca:
This is a feature request to extend the `stream_join` IP or provide the extended functionality in a separate IP.

Currently, the `stream_join` IP enforces:
1. all of the streams to be joined together
2. all input handshakes to occur simultaneously, i.e. an input stream is granted ready only after all input streams are valid

Conversely, we seek the following features:
1. dynamically select a subset of streams to be joined
2. allow input streams to be granted independently of each other

Feature (1) is similar to the functionality `stream_fork_dynamic` provides over `stream_fork`. The `stream_fork` IP is "specular" in its functionality to `stream_join`.

PR #151, opened 2022-08-12
Add stream_join_dynamic IP

colluca:
This PR addresses feature request (1) from issue #150.

### DIFF
--- a/src_files.yml
+++ b/src_files.yml
@@ -37,6 +37,7 @@ common_cells_all:
     - src/stream_fork.sv
     - src/stream_intf.sv
     - src/stream_join.sv
+    - src/stream_join_dynamic.sv
     - src/stream_mux.sv
     - src/stream_throttle.sv
     - src/sub_per_hash.sv
--- a/Bender.yml
+++ b/Bender.yml
@@ -53,7 +53,7 @@ sources:
   - src/stream_filter.sv
   - src/stream_fork.sv
   - src/stream_intf.sv
-  - src/stream_join.sv
+  - src/stream_join_dynamic.sv
   - src/stream_mux.sv
   - src/stream_throttle.sv
   - src/lossy_valid_to_stream.sv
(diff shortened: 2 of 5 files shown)


niwis:
@colluca thank you for this contribution! As you mention in #150, the code of `stream_join` and `stream_join_dynamic` is very similar. Would it be possible to merge them? e.g. by modifying `stream_join` to wrap `stream_join_dynamic` and setting `sel_i = '1`. Or (perhaps even better) by instantiating `stream_join` in `stream_join_dynamic` and setting `.inp_valid_i(inp_valid_i | ~sel_i)`

colluca:
> @colluca thank you for this contribution! As you mention in #150, the code of `stream_join` and `stream_join_dynamic` is very similar. Would it be possible to merge them? e.g. by modifying `stream_join` to wrap `stream_join_dynamic` and setting `sel_i = '1`. Or (perhaps even better) by instantiating `stream_join` in `stream_join_dynamic` and setting `.inp_valid_i(inp_valid_i | ~sel_i)`

Sure :) I would favour the first option, since `stream_join_dynamic` is a superset of the functionality of `stream_join`. And I believe it only makes sense to keep `stream_join` for the reduced number of ports, correct?

niwis:
The interface of `stream_join` should not change in either case. The second option would be structurally closer to the `stream_join` logic and less invasive, but I'm okay with both

colluca:
I realized there's a corner case (when the mask disables all incoming streams) that would maybe need to be handled differently. I believe the desired behavior is not to merge anything, i.e. not to forward any valid. This makes it a bit more difficult (or inefficient if not optimized by the synthesis tool) to base `stream_join_dynamic` on `stream_join` rather than the opposite.

I pushed a first proposal where I base `stream_join` on `stream_join_dynamic`, which we can of course discuss further.

src_files.yml
```diff
@@ -37,6 +37,7 @@ common_cells_all:
     - src/stream_fork.sv
     - src/stream_intf.sv
     - src/stream_join.sv
+    - src/stream_join_dynamic.sv
```
niwis, line 40: `stream_join` now depends on `stream_join_dynamic`, so the former should go to level 1 (same for `Bender.yml`)

niwis (changes requested):
Thank you @colluca. If this is ready from your side, could you rebase onto master to re-run CI? Then, this should be ready to be merged.

niwis (approved):
Thanks a lot!

PR #153, opened 2022-08-22
Allow out-of-bounds top end address in addr_map

micprog:
This allows for use of an address map where the top end address is zero, indicating the end of the address space (or wrap around as there are no longer any bits available).


niwis (approved):
LGTM!

PR #160, opened 2022-11-25
`stream_fifo_optimal_wrap`: Remove assertions

fischeti:
In my opinion, the `Full push` and the `Empty pop` assertions are unnecessary in the `stream_fifo_optimal_wrap`. Empty pops and Full pushes are prevented inside the modules anyway.

Those asserts for example require that the `ready` can only be asserted when the `valid` is asserted, which is an unnecessary restriction in my opinion. The AXI ready signals e.g. are asserted independently of the valid signals.

What do you think @thommythomaso?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - Allow out-of-bounds (i.e. `'0`) top end address in addr_map of `addr_decode` module for end of address space.
 - Update CI.
+- Remove asserts in `stream_fifo_optimal_wrap`
 
 ## 1.25.0 - 2022-08-10
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
 and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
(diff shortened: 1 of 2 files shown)


thommythomaso:
> In my opinion, the `Full push` and the `Empty pop` assertions are unnecessary in the `stream_fifo_optimal_wrap`. Empty pops and Full pushes are prevented inside the modules anyway.
> 
> Those asserts for example require that the `ready` can only be asserted when the `valid` is asserted, which is an unnecessary restriction in my opinion. The AXI ready signals e.g. are asserted independently of the valid signals.
> 
> What do you think @thommythomaso?

You can remove the assertions.

thommythomaso (approved):
:+1: LGTM

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.
 ### Changed
 - Allow out-of-bounds (i.e. `'0`) top end address in addr_map of `addr_decode` module for end of address space.
 - Update CI.
+- Remove asserts in `stream_fifo_optimal_wrap`
```
niwis, line 14: This is not included in `v1.26.0`. Could you add this to a new "unreleased" section? i.e.
```
## Unreleased
### Changed
- Remove asserts ...

## 1.26.0 ...
```

niwis (changes requested):

CHANGELOG.md
fischeti: Uups, sorry. I changed it now.

niwis (approved):
Thanks! LGTM

PR #162, opened 2022-12-06
fall_through_register: Remove superfluous `$size` call to allow the use of union types in DCnxt 2022.03

thommythomaso:


niwis (approved):
Thanks, LGTM

PR #167, opened 2023-02-14
Add priority encoder

emanueleparisi:
Hi! I designed this priority encoder in the context of a project we are carrying on at University of Bologna, about enhancing the CVA6 core with security features. 

This is a fully-combinational, generic, priority encoder that takes as input a `N_BITS` signal and returns the index of the first bit set, giving higher priority to the LSB bit, in case multiple bits are set. If the input is zero, a output flag is cleared. Any input signal whose width is bigger than 1 is legal.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -43,6 +43,7 @@ sources:
   - src/onehot_to_bin.sv
   - src/plru_tree.sv
   - src/popcount.sv
+  - src/priority_encoder.sv
   - src/rr_arb_tree.sv
   - src/rstgen_bypass.sv
   - src/serial_deglitch.sv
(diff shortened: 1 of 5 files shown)


niwis:
Hello @emanueleparisi, thanks a lot for your contribution! Do you think the same functionality could be achieved with the [lzc](https://github.com/pulp-platform/common_cells/blob/master/src/lzc.sv) module, by setting `MODE` to `0` (trailing zero)?

emanueleparisi:
Hi @niwis, thank you for you comment! Actually yes, from the functional point of view `lzc` serves the purpose. Also, I compared the area of the two modules running a couple of synthesis on Vivado 2021 (OptArea-High), and the `lzc` module seems slightly better than the one I developed. I guess the best thing to do is to close the PR. Thank you for your time!

niwis:
Okay, thanks a lot for looking into this!

PR #168, opened 2023-02-15
registers: Update load-enable FF macros

stmach:
Removes the ternary operator from load-enable FF macros as this causes synthesis of a physical feedback path from the current state `__q` to the next state `__d` in synthesis.


niwis (approved):
Thanks a lot! LGTM

PR #171, opened 2023-03-08
Allow clock divider configuration while clock is disabled

meggiman:
The FSM of the configurable clock divider contains an unnecessary gating condition to only allow divider configuration while the output clock is enabled. Although configuration during enabled output clock is perfectly safe since the module temporarily gates itself to avoid any output glitches, it might still be desirable from an application point of view to disable the output clock for a longer time period and re-enable it with a different clock divider value. With the current implementation this is not possible since at least one or two clock edges of the old clock period might still escape the module before the reconfiguration with enabled output clock takes effect.

This PR fixes this and adds a new test to the TB to check if the divider can be configured correctly while `en_i` is de-asserted. Thanks to @fimtrey for pointing out this issue.

### DIFF
--- a/src/clk_int_div.sv
+++ b/src/clk_int_div.sv
@@ -139,7 +139,7 @@ module clk_int_div #(
       IDLE: begin
         gate_en_d     = 1'b1;
         toggle_ffs_en = 1'b1;
-        if (en_i && div_valid_i) begin
+        if (div_valid_i) begin
           if (div_i == div_q) begin
             div_ready_o      = 1'b1;
           end else begin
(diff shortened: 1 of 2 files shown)


fimtrey (commented):
As indicated it's allowed to tie div_valid_i to 1. I've the `div_i` is 0 at any moment, an issue occurs.

If `div_q = 0` and `en_i = 0`, it will never be able to reset the state to IDLE because the following condition can never be true [(line 187)](https://github.com/pulp-platform/common_cells/blob/eef642789e4ed89c9b5590ec1f6477d502a079bf/src/clk_int_div.sv#L187)
``` systemverilog
    if (cycle_cntr_q == div_q - 1) begin
```

niwis (approved):
LGTM, thanks a lot!

niwis:
@fimtrey as the condition you are describing is orthogonal to this PR, could you move it to an issue?

PR #177, opened 2023-04-13
Fix HideStrb feature of mem_to_banks

micprog:
With this change, mem_to_banks assumes at least 1 cycle latency to the memory bank, but cuts a possible timing loop (e.g., when in axi_to_mem).


niwis (approved):
LGTM

PR #180, opened 2023-04-17
Add multiaddr_decode IP

colluca:
This IP provides address-based multicasting capabilities. By providing a mask together with the address, we can compactly represent a set of addresses to compare with the address map.

A question might be if we want to merge this functionality in the `addr_decode` module or keep it separate as is the case now. Integration of the `addr_decode_napot` with `addr_decode` was already a step in this direction, as the prior uses the same encoding for the address map as the `multiaddr_decode` module.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -69,6 +69,7 @@ sources:
   - src/cdc_2phase.sv
   - src/cdc_4phase.sv
   - src/addr_decode.sv
+  - src/multiaddr_decode.sv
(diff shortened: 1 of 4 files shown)


niwis (changes requested):
I think it's OK to add this as a new module. Could you update src_files.yml and README.md and rebase?

niwis (approved):
Cheers!

PR #182, opened 2023-05-08
Fix deadlock on div_i == 0 and avoid hold issues on ungated_output_clock

### LINKED ISSUE #172: Clock divider configuration fails if divider value is 0
@meggiman I ran into another issue with the clock divider.

Following the description 
> // It is thus safe to statically tie the valid signal to logic high if
> // we can guarantee, that the div_i value remains stable long enough (upper
> // limit 2 output clock cycles).

I tied `valid_i` to 1 and connected `div_i` to a register which has a reset value 0. 

That's when I noticed that after `div_q = 0` is loaded, I'm not able to reconfigure the clock divider. 

As far as I understood, it hangs in the `WAIT_END_PERIOD` state, because the following condition can never be satisfied.
https://github.com/pulp-platform/common_cells/blob/f9528fa6e4d3080605c01074bdaf62dcb4aff7d2/src/clk_int_div.sv#L187

Thanks @niwis for pointing out that this is indeed independent of the configuration while the divider is turned off, it just popped up checking https://github.com/pulp-platform/common_cells/pull/171.

meggiman:
Instead of storing div_i directly in the internal register div_q we normalize the value to be strictly larger than 0. This avoids the deadlock issue when preloading with 0. Furthermore this commit removes all combinational checks on the phase of `ungated_output_clock` to avoid hold issues in synthesis. This checks were required to wait until the clock enable signal has been latched by the ICG. Now, we instead use a dedicated flip-flop that tracks the current status of the output clock gate.

This fixes #172.

### DIFF
--- a/lint/common_cells.style.waiver
+++ b/lint/common_cells.style.waiver
@@ -4,9 +4,9 @@ waive --rule=typedef-structs-unions --line=29 --location="src/ecc_encode.sv"
 # That is a known issue with string parameter in Synopsys DC
 waive --rule=explicit-parameter-storage-type --line=19 --location="src/stream_arbiter.sv"
 waive --rule=explicit-parameter-storage-type --line=19 --location="src/stream_arbiter_flushable.sv"
-waive --rule=always-ff-non-blocking --line=262 --location="src/clk_int_div.sv"
-waive --rule=always-ff-non-blocking --line=265 --location="src/clk_int_div.sv"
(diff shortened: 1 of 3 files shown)


fimtrey:
I tested in on a pynq-z2 FPGA and it works for my case and there're no hold issues anymore. Thank you @meggiman !

niwis (approved):
Thanks! LGTM

PR #183, opened 2023-05-09
Clk int div static

meggiman:
This PR adds a convenience wrapper for the `clk_int_div` module that ties of the inputs to provide a static clock divider with a fixed division value. The PR also contains a TB for this new module that verifies the clock divider for various different parametrizations.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -26,7 +26,7 @@ stages:
   script:
     - bender script vsim -t test > compile.tcl
     - $VSIM -c -quiet -do 'source compile.tcl; quit'
-    - $VSIM -c $TOPLEVEL -do "run -all" $PARAM1 $PARAM2
+    - $VSIM -c $TOPLEVEL -voptargs="-timescale 1ns/100ps" -do "run -all" $PARAM1 $PARAM2
     - (! grep -n "Error:" transcript)
 
 tests:
@@ -38,6 +38,7 @@ tests:
           - cb_filter_tb
           # - cdc_fifo_clearable_tb
           - clk_int_div_tb
+          - clk_int_div_static_tb
           - clk_mux_glitch_free_tb
           - id_queue_tb
           - isochronous_crossing_tb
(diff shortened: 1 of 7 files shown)


niwis (commented):
@meggiman thank you for this contribution! I just have one thought: this module has almost the same spec as [`clk_div`](https://github.com/pulp-platform/common_cells/blob/master/src/deprecated/clk_div.sv). Would it make sense to replace `clk_div` with this wrapper to a) avoid code duplication and b) prevent people from accidentally using the (broken) `clk_div`?

meggiman:
This is a valid input. The only issue I could see with this approach is, that there might be IPs that rely on the faulty duty cycle behavior of the old clk_div as a feature. What do you think @bluewww?

bluewww:
The "problem" with the `clk_div` module is that it has a skewed duty cycle since it basically let's through one short clock pulse for every `RATIO` clock ticks. I want to mention here that the idea to have one clock pulse being propagated every N cycles is totally viable (and being done in real-life) unfortunately `clk_div` does this it the wrong way. The correct way to do this would be to use a down counter and a clock gate latch. Some people sometimes refer to that design as "punch through divider".

Now whether it is a good idea to replace this skewed (and buggy divider withouth my patch I believe)  with a balanced divider I don't know. It might be better to problem implement a punch through divider and replace it with that.

I have been using a patched version of `clk_div` but I will switch to `clk_int_div` myself.

niwis:
I honestly would prefer to replace `clk_div` with this module and un-deprecate it. I don't think the skewed duty cycle or the [exponential divisor](https://github.com/pulp-platform/common_cells/pull/185#discussion_r1197386742) are intended features of `clk_div` or that anyone relies on them (they aren't even documented). And its implementation has more flaws, so I think this would be the safest approach. Of course, we could still add a module `clk_div_skew` (or similar) that would skew the clock explicitly.

But I don't want to stall this for much longer. If both of you agree that we should keep the old `clk_div` as is, I'm also okay with that.

meggiman:
I have no strong opinion here. But if we change the existing clk_div module it has to be released in a major version bump since it will change the behaviour of an existing module (and the interface).

PR #185, opened 2023-05-09
Fixes and Extensions for deprecated modules

bluewww:
* deprecated/clk_div.sv: IP was not working at all
* deprecated/clock_divider.sv: Make division value readable
* src/stream_delay: Allow larger counts

### DIFF
--- a/src/popcount.sv
+++ b/src/popcount.sv
@@ -27,7 +27,7 @@ module popcount #(
    localparam int unsigned PaddedWidth = 1 << $clog2(INPUT_WIDTH);
 
    logic [PaddedWidth-1:0]           padded_input;
-   logic [PopcountWidth-2:0]         left_child_result, right_child_result;
+   logic [PopcountWidth-1:0]         left_child_result, right_child_result;
 
    //Zero pad the input to next power of two
    always_comb begin


src/stream_delay.sv
```diff
@@ -44,7 +44,7 @@ module stream_delay #(
         state_e state_d, state_q;
 
         logic       load;
-        logic [3:0] count_out;
+        logic [CounterBits:0] count_out;
```
alex96295, line 47: I think here is `[CounterBits-1:0]`

because `count_out` is `q_o` in the `counter` module, and has width `WIDTH` (https://github.com/pulp-platform/common_cells/blob/control-pulp/src/counter.sv#L25)

So, it should be the same width of `counter_load`

src/deprecated/clock_divider.sv
```diff
@@ -53,6 +53,7 @@ module clock_divider
     input  logic [7:0] clk_div_data_i,
     input  logic       clk_div_valid_i,
     output logic       clk_div_ack_o,
+    output logic [7:0] clk_div_data_o,
```
niwis, line 56: I would prefer not to change the interface of a deprecated module. Is this necessary?

src/deprecated/clk_div.sv
niwis, line 1: Thanks. There are several bugs in this module (#46) and I would propose to make this a wrapper for `clk_int_div` to prevent anyone from using this broken version (as discussed in #183).

niwis: As far as I can tell, the changes here also change the behaviour (effectively by changing the divisor from `2^RATIO` due to the overflow to `RATIO`, or am I missing something?) And the duty cycle is still off. My opinion: either we leave the behaviour unchanged (i.e. broken) or fix this properly. I'm okay with both, but doing something in between might give a false sense of security.

src/deprecated/clock_divider.sv
bluewww: This is just a feature I needed. I don't know any users of those module. The UDMA has a fork of this divider within its own codebase.

src/deprecated/clk_div.sv
bluewww: I don't think it divides by `2^RATIO`. I would argue that `RATIO[RATIO-1:0]` equals `RATIO` for `RATIO > 0`. I think the `RATIO[RATIO-1:0]` expression is just nonsense. Assume `RATIO=5`. In that case it means to the lower 5 bits from the constant 5 which is a 5 bit value containing 5.

niwis: I agree with that part. However, `counter_q` is currently not cleared when reaching `RATIO`. I would therefore argue that it will increment all the way to `RATIO^2-1` (since `counter_q` is currently `RATIO` bits wide) before overflowing and resuming from `'0`. During these `RATIO^2` input-clock cycles, there is exactly one output-clock pulse when `counter_q == RATIO`.

src/deprecated/clock_divider.sv
niwis: I'm not a big fan of maintaining deprecated modules, since they were usually deprecated for a good reason and should no longer be used.
However, if you need this, I propose a safe approach by creating a new module (`clock_divider_observable` or similar) with this functionality and changing `clock_divider` to wrap this new module, not exposing `clk_div_data_o`.

src/popcount.sv
```diff
@@ -27,7 +27,7 @@ module popcount #(
    localparam int unsigned PaddedWidth = 1 << $clog2(INPUT_WIDTH);
 
    logic [PaddedWidth-1:0]           padded_input;
-   logic [PopcountWidth-2:0]         left_child_result, right_child_result;
+   logic [PopcountWidth-1:0]         left_child_result, right_child_result;
```
niwis, line 30: as far as I can tell, the parent's result is one bit wider than that of the children, so the old version looks correct to me. @bluewww @meggiman can you confirm that this should be merged?

src/stream_delay.sv
niwis: Fixed in #190.

src/popcount.sv
bluewww: This is just a hacky workaround right now for myself. I think @meggiman rewrote the popcount to use regular sv constructs which should be used in the future.

Withouth my hack the bug is buggy though for the case since it will result in a `logic [-1:0]` vector

bluewww: See https://github.com/pulp-platform/common_cells/blob/master/src/popcount.sv

niwis: Right! Resolved in #188, then.

niwis:
Superseded by #188, #190

PR #188, opened 2023-05-24
Simplify popcount and add proper support for all input widths

### LINKED ISSUE #187: popcount for INPUT_WIDTH=1
`popcount.sv` contradicts itself in the description and this commit https://github.com/pulp-platform/common_cells/commit/bf4db7cfa01030f36a59c1a4d21ee233f6aee01e whether the case `INPUT_WIDTH=1` is supported.

Furthermore, it seems like the https://github.com/pulp-platform/common_cells/commit/bf4db7cfa01030f36a59c1a4d21ee233f6aee01e doesn't properly add full support for `INPUT_WIDTH=1` since 

```verilog
logic [PopcountWidth-2:0]         left_child_result, right_child_result;
```

(from here https://github.com/pulp-platform/common_cells/blob/master/src/popcount.sv#L30)

becomes negative causing at least VCS to crash. A simple hacky fix is to change the `-2`  to `-1` but I'm not sure if that works out well.

Maybe @meggiman can you comment on this (since you seem to know the most being the designer of this module) ?

@alex96295 fyi

meggiman:
The popcount implementation so far used a recursive implementation using a binary adder tree. However, experiments using reasonably modern synthesis tools show that the logic optimization heuristics achieve the same or better results in less compilation time when the popcount logic is described in a flat for loop instead of nested modules.
While simplifying the popcount implementation, this change also fixes #187

### DIFF
--- a/src/popcount.sv
+++ b/src/popcount.sv
@@ -11,50 +11,30 @@
 // Author: Manuel Eggimann <meggimann@iis.ee.ethz.ch>
 
 // Description: This module calculates the hamming weight (number of ones) in
-// its input vector using a balanced binary adder tree. Recursive instantiation
-// is used to build the tree.  Any unsigned INPUT_WIDTH larger or equal 2 is
-// legal.  The module pads the signal internally to the next power of two.  The
-// output result width is ceil(log2(INPUT_WIDTH))+1.
+// its input vector. Any unsigned INPUT_WIDTH larger or equal 1 is legal. The output result
+// width is ceil(log2(INPUT_WIDTH))+1.
+//
+// This module usded tobe implemented using a binary added tree. However,
+// the heuristics of modern logic Synthesizers work much better with a flat high
+// level description using a for loop and yield exactly the same or even better results.
+
 
 module popcount #(
-    parameter int unsigned INPUT_WIDTH = 256,
-    localparam int unsigned PopcountWidth = $clog2(INPUT_WIDTH)+1
+    parameter  int unsigned INPUT_WIDTH   = 256,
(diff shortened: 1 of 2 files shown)


bluewww (approved):
LGTM!

src/popcount.sv
```diff
@@ -11,50 +11,30 @@
...
 
 // Description: This module calculates the hamming weight (number of ones) in
-// its input vector using a balanced binary adder tree. Recursive instantiation
-// is used to build the tree.  Any unsigned INPUT_WIDTH larger or equal 2 is
-// legal.  The module pads the signal internally to the next power of two.  The
-// output result width is ceil(log2(INPUT_WIDTH))+1.
+// its input vector. Any unsigned INPUT_WIDTH larger or equal 1 is legal. The output result
+// width is ceil(log2(INPUT_WIDTH))+1.
+//
+// This module usded tobe implemented using a binary added tree. However,
```
niwis, line 17: fix typo
```suggestion
// This module used to be implemented using a binary added tree. However,
```

test/popcount_tb.sv
```diff
@@ -31,6 +31,9 @@ end while (0)
 module popcount_tb;
 
    //---------------- Signals connecting to MUT ----------------
+   logic data_w1;
+   logic  popcount_w1;
```
niwis, line 35: fix alignment
```suggestion
   logic popcount_w1;
```

niwis (changes requested):
Thanks a lot! Just one typo from my side

niwis (approved):

PR #189, opened 2023-05-24
Lossy valid to stream

meggiman:
This module helps to deal with sources that use a valid-only interface, that
 is they do not support backpressure i.e. cannot handle the case where the
 sink is not ready to accept a value. The module is implemented as FIFO with 2
 elements. In contrast to a regular FIFO that stops accepting new transactions
 once the FIFO is full, this IP overwrites the last element entered into the
 FIFO. This means the input is always ready to accept new transactions,
 however, intermediate transactions might be overwritten by the latest one. On
 the output side, the module behaves like a regular ready-valid source i.e.
 once valid is asserted, data_o remains stable until the sink consumes them
 (by asserting ready_i).

The `lossy_valid_to_stream` module is helpful to connect configuration registers with
 IPs that could cause back pressure. In this case we might not care how long
 it takes for the config value to be sent to the IP but if we change the
 config value we want the latest value to be used regardless whether the
 previous value has already been forwarded or not.


src/lossy_valid_to_stream.sv
```diff
@@ -0,0 +1,139 @@
...
+      data_o = data_i;
+    end else if (valid_o && ready_i) begin
+      read_ptr_d = read_ptr_q + 1'b1;
+    end
+  end
+
+  always_comb begin: count_transactions
+    pending_tx_counter_d = pending_tx_counter_q;
+    if (valid_i && valid_o && ready_i) begin
+      pending_tx_counter_d = pending_tx_counter_q;
```
niwis, line 111: I guess this is redundant

niwis (commented):
Looks good to me in general, thanks a lot!

I have one question: The documentation mentions the case that *...we want the latest value*. If I understand correctly, the module needs to be read twice to obtain the newest value (the first value read will be the *oldest* uncommited value, since the signals are kept stable). Is that correct?

One more comment inline.

src/lossy_valid_to_stream.sv
meggiman: Indeed it is. That was a voluntarily decision to make the three cases (push, pop and simultaneous push and pop) a bit more clear in the code.

meggiman:
Yes that is correct. Since we are not allowed to revoke an already started transaction that is the only option we have if I am not mistaken.

niwis (approved):
Okay, clear! Thanks a lot

Issue #191, opened 2023-07-24
Width mismatch error in assign statement with VCS strict compile


rsharanesh:
In the below assignment, the RHS being a genvar is 32-bits wide, while the LHS which is parameterized and need not be always of 32-bits wide. Thus the compiler throws a width mismatch error. 

https://github.com/pulp-platform/common_cells/blob/0989ff73d0315922791bf42137c0ce0cbb4a76ca/src/onehot_to_bin.sv#L26

Can the code be modified to be generic enough with the assignment like this,
``assign tmp_i = i[BIN_WIDTH-1:0];``

PR #192, opened 2023-07-24
Fix width mismatch assignment issue #191

rsharanesh:
As mentioned in the issue, this would make the assignment more generic and compile clean with strict compile in few simulators.

### DIFF
--- a/src/onehot_to_bin.sv
+++ b/src/onehot_to_bin.sv
@@ -23,7 +23,7 @@ module onehot_to_bin #(
         logic [ONEHOT_WIDTH-1:0] tmp_mask;
             for (genvar i = 0; i < ONEHOT_WIDTH; i++) begin : gen_il
                 logic [BIN_WIDTH-1:0] tmp_i;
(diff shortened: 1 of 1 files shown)


src/onehot_to_bin.sv
```diff
@@ -23,7 +23,7 @@ module onehot_to_bin #(
         logic [ONEHOT_WIDTH-1:0] tmp_mask;
             for (genvar i = 0; i < ONEHOT_WIDTH; i++) begin : gen_il
                 logic [BIN_WIDTH-1:0] tmp_i;
-                assign tmp_i = i;
+                assign tmp_i = i[BIN_WIDTH-1:0];
```
niwis, line 26: I guess the proper way to handle this would be 
```suggestion
                assign tmp_i = BIN_WIDTH'(i);
```
What do you think? Could you check if this would solve your issue?

niwis (changes requested):
Thanks a lot for spotting this. I have one suggestion regarding the solution.

niwis (approved):

PR #193, opened 2023-07-26
unread: Assign do dummy net when synthesizing with vivado

huettern:
Synthesizing with Vivado results in the module `unread` being a black box and subsequent implementation fails.

This patch proposes to assign the input to a fake net to mitigate this error.


niwis (approved):
Thanks a lot!

PR #194, opened 2023-07-27
Add `mem_to_banks_detailed`

micprog:
Adds sideband data in the response path to be managed externally. `mem_to_banks` now wraps the `mem_to_banks_detailed`.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -98,9 +98,11 @@ sources:
   # Level 3
   - src/cdc_fifo_gray_clearable.sv
   - src/cdc_2phase_clearable.sv
-  - src/mem_to_banks.sv
+  - src/mem_to_banks_detailed.sv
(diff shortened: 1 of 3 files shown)


glaserf (commented):
I second what @niwis has said. Looks good to me.

One suggestion would be to generalize the error-response channel to generic side-channel memory response info by renaming it accordingly, but I'm not sure whether such use cases (besides error responses) do exist.

micprog:
The updated `mem_to_banks` wrapper was tested with the [`axi_to_mem_banked`](https://github.com/pulp-platform/axi/blob/master/test/tb_axi_to_mem_banked.sv) testbench in the AXI repository, it still works as originally intended.

niwis (approved):
Thanks a lot, LGTM

PR #195, opened 2023-08-08
Set bank count to nonzero in default parameter to avoid div by 0

uge:
Verilator crashes parsing mem_to_banks.sv because of a divide by zero if NumBanks is 0


niwis (changes requested):
Thanks a lot @uge. There were recent changes to `mem_to_banks`. Could you rebase your patch onto `master` and add the same change to `mem_to_banks_detailed`?

niwis (approved):
Awesome, thank you!

PR #196, opened 2023-08-09
Fix connection as mem_to_banks localparams should not be overwritten

micprog:


alex96295 (approved):

niwis (approved):
LGTM. Do you need a version bump for that?

alex96295:
> LGTM. Do you need a version bump for that?

If possible, yes please

PR #201, opened 2023-10-20
Upstream fixes

asmolenskiy:
"pragma" prefix for translate_on/off might be considered as invalid by many tools w/o additional settings

by default valid prefix for translate_on/off is "synopsys", and both "pragma" and "synopsys" are valid for synthesis_on/off


niwis:
Closed in favor of #210.

PR #206, opened 2023-12-14
plru_tree: Assert output is onehot

michael-platzer:
@niwis Another thing that just came to my mind: it might make sense to add an assertion to verify that the output is always onehot.

### DIFF
--- a/src/plru_tree.sv
+++ b/src/plru_tree.sv
@@ -123,6 +123,10 @@ module plru_tree #(
     initial begin
         assert (ENTRIES == 2**LogEntries) else $error("Entries must be a power of two");
     end
+
+    output_onehot : assert property(
+        @(posedge clk_i) disable iff (~rst_ni) ((plru_o & (plru_o - 1)) == '0))
+        else $fatal (1, "More than one bit set in PLRU output.");
 `endif
 // pragma translate_on
 
--- a/src/plru_tree.sv
+++ b/src/plru_tree.sv
@@ -125,7 +125,7 @@ module plru_tree #(
     end
 
     output_onehot : assert property(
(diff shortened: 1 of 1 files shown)


src/plru_tree.sv
```diff
@@ -123,6 +123,10 @@ module plru_tree #(
     initial begin
         assert (ENTRIES == 2**LogEntries) else $error("Entries must be a power of two");
     end
+
+    output_onehot : assert property(
+        @(posedge clk_i) disable iff (~rst_ni) ((plru_o & (plru_o - 1)) == '0))
```
niwis, line 128: For better readability, can we use the corresponding system function?
```suggestion
        @(posedge clk_i) disable iff (~rst_ni) ($onehot0(plru_o))
```

niwis (commented):
Makes sense! Just one question

src/plru_tree.sv
michael-platzer: Sure, thanks for making me aware of that :slightly_smiling_face:

src/plru_tree.sv
```diff
@@ -123,6 +123,10 @@ module plru_tree #(
     initial begin
         assert (ENTRIES == 2**LogEntries) else $error("Entries must be a power of two");
     end
+
+    output_onehot : assert property(
+        @(posedge clk_i) disable iff (~rst_ni) ($onehot0(plru_o))
```
niwis, line 128: Missing closing parenthesis, my bad
```suggestion
        @(posedge clk_i) disable iff (~rst_ni) ($onehot0(plru_o)))
```

niwis (changes requested):

src/plru_tree.sv
michael-platzer: Sorry, didn't check either.

PR #210, opened 2024-01-21
Renamed `translate_off` to `` `ifndef SYNTHESIS ``

sifferman:
Hello!

According to the IEEE 1364.1-2005 spec 6.3.2, the use of `translate_off`/`translate_on` is discouraged as opposed to `` `ifndef SYNTHESIS ``. This change gives support to newer SystemVerilog frontends such as sv2v, Verible, and PySlint.

Thanks!

### DIFF
--- a/src/addr_decode_dync.sv
+++ b/src/addr_decode_dync.sv
@@ -124,7 +124,7 @@ module addr_decode_dync #(
   // Assumptions and assertions
   `ifndef COMMON_CELLS_ASSERTS_OFF
   `ifndef XSIM
-  // pragma translate_off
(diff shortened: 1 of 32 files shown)


niwis (changes requested):
@sifferman thanks a lot for your contribution, and sorry for my late response. To make this complete, could you also replace the pragmas in [stream_fifo_optimal_wrap.sv](https://github.com/sifferman/common_cells/blob/master/src/stream_fifo_optimal_wrap.sv)?

I'm open to updating this in the deprecated modules but don't think it's strictly necessary.

sifferman:
Thanks!

PR #214, opened 2024-03-05
Added macro with asynchronous reset and synchronous clear.

yvantor:


yvantor:
@niwis not putting this in draft state as I think it will not need much more than this, but I think it might make sense to wait for the activity on CVA6 to complete before evaluating a merge.

niwis (approved):
@yvantor thanks a lot, LGTM. I'm about to make a new release, so I think we can merge this already and follow up if needed.

PR #217, opened 2024-03-12
`stream_omega_net`: Fix non-existing output assertion

colluca:
Fixes the same problem described in https://github.com/pulp-platform/common_cells/pull/204, but for the `stream_omega_net` IP.

This urges a new release, as v1.33.0 broke certain IPs (RTL simulation of the Snitch cluster) which previously worked with v1.32.0.


niwis (approved):
LGTM

PR #219, opened 2024-04-02
stream-xbar: Add payload stability assertion mask

paulsc96:
Fixing the reset polarity of assertions in `stream_xbar` and `stream_omega_net` (#200) uncovered numerous assertion failures in our  designs using `common_cells`.  One assertion failure we observe may result from intentional, functionally correct design.  

`stream_xbar` and `stream_omega_net` are commonly used for in-cycle banked SRAM interconnects (e.g. PULP TCDM), wherein parts of a valid payload may freely change if they are not relevant to that payload.

The most common example is a shared request channel for reads and writes. The write data is always part of the payload, but on reads (i.e. when the write enable in the payload is low), it is irrelevant to the request. Thus, efficient designs may opt to hardwire write data to a write-specific datapath, which may result in write data changes while a valid read request is pending. 

While functionally harmless, such changes trigger the current data stability assertions in `stream_xbar` and `stream_omega_net` which provide either strict stability assertions on all payload bits (`AxiVldRdy == 1`) or no stability assertions at all  (`AxiVldRdy == 0`).

This MR adds a mask parameter `AxiVldMask` allowing designers to choose on which payload bits stability should be asserted. Since the default mask is `'1`, this change is **non-breaking**.

### DIFF
--- a/src/stream_omega_net.sv
+++ b/src/stream_omega_net.sv
@@ -38,6 +38,10 @@ module stream_omega_net #(
   /// When this is set, valids have to be asserted until the corresponding transaction is indicated
   /// by ready.
   parameter int unsigned LockIn      = 1'b1,
+  /// If `AxiVldReady` is 1, which bits of the payload to check for stability on valid inputs.
+  /// In some cases, we may want to allow parts of the payload to change depending on the value of
+  /// other parts (e.g. write data in read requests), requiring more nuanced external assertions.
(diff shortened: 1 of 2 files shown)


colluca (approved):
LGTM :)

bluewww:
Sensible change lgtm.

niwis (approved):
Thanks, LGTM

PR #223, opened 2024-05-15
Fix else statement in FFARNC.

yvantor:


niwis (approved):
Critical fix, thanks!

PR #224, opened 2024-07-11
`mem_to_banks_detailed`: Ensure no spurious response after full dead write

micprog:


niwis (commented):
@micprog thanks for the contribution! Three points from my side:
1. Masking `dead_response` with `~dead_write_fifo_empty` makes sense to me.
2. Is the reduction (lines 156, 159, 161) really necessary?
3. Why do we need to change the input of the `dead_read_fifo` from `bank_oup` to `bank_req`?

PR #226, opened 2024-08-20
Add passthrough mode for mem_to_banks with NumBanks==1

micprog:


niwis (commented):
Thanks @micprog! I guess that changing the functional behaviour of `mem_to_banks_detailed` for `NumBanks == 1` would require a major release? Also, I think we should somehow handle conflicting parametrisations (e.g. `NumBanks == 1 && HideStrb == 1`). Perhaps it would be cleaner to change this in the instantiating module and just add a warning here if `NumBanks == 1`. What do you think?

phsauter:
I have been thinking about this and the correct solution to me seems to be warning and assertions.  
This change could potentially introduce critical paths due to the FIFO removal and it functionally changes what it does in the case of back-pressure.  
So in my opinion this should be considered closed and we should add the warnings instead (I will leave it open until this is done).

phsauter:
The asserts/warnings will be in v2 with #299 , the rest is dropped. I will archive the branch.

PR #233, opened 2024-09-18
Replace all asserts & assumes with macros from `assertions.svh`

### LINKED ISSUE #232: Why are the common cells not using the assertion macros?
Is there a particular reason why the modules in this repository are not using the assertion macros in [`assertions.svh`](https://github.com/pulp-platform/common_cells/blob/master/include/common_cells/assertions.svh)?

Using these macros seems to make sense, as it allows more freedom in defining custom ways to report assertions (via the `` `ASSERT_RPT`` macro) and also to override macros that typically disable assertions such as `` `SYNTHESIS`` by forcefully defining `` `INC_ASSERT``.

@niwis Would a patch that replaces all `assert` (and `assume`) statements in this repository with macros from `assertions.svh` be accepted? Is there anything particular that you would like to have addressed as part of such a change?

michael-platzer:
This PR replaces all assertions (and assumes) in the common cells sources with macros from [`assertions.svh`](https://github.com/pulp-platform/common_cells/blob/master/include/common_cells/assertions.svh).

Additionally, the `assertions.svh` itself is modified as follows:
- Some defines (such as the guard block at the start of the header) are renamed for consistency and to avoid name collisions in the same style as in [`registers.svh`](https://github.com/pulp-platform/common_cells/blob/master/include/common_cells/registers.svh).
- The `` `ifndef VERILATOR`` is removed because Verilator has supported assertions for a while now.
- An option to override any of the default defines that cause assertions to be turned off (such as `SYNTHESIS`) is added (so even if a tool defines, e.g., `SYNTHESIS`, it still has the option to forcefully enable assertions by also defining `ASSERTS_OVERRIDE_ON` now — yes, there are tools out there that require that :roll_eyes:).
- Helper macros are undefined at the end of the file (a comment did claim that this was already the case, but it was not).
- An optional extra argument is added to all assertions macros, which allows to specify a custom error message to be displayed when the assertion fails. For backwards compatibility, this extra argument is added at the end and defaults to  `""`.

In order to replace all the assertions in the sources while avoiding typos and other oversights as best as possible, I have replaced almost all of them using sed commands. The specific commands used for each change are part of the respective commit messages.

Some noteworthy subtle changes that go along with this move:
- The macros in `assertions.svh` require that all assertions have a name, so I added one for all those who did not have one (and I tried to hard to keep them consistent).
- So far, assertions used a mix of `$fatal`, `$error`, and sometimes `$warning` to report failing assertions. The macros in `assertions.svh` universally use `$error` and all instances of `$fatal` are now changed to `$error`. I reviewed all those that used `$warning` but since they all appear to be pretty critical issues as well, they now also use the macros which report an `$error` (the affected files are `src/addr_decode_dync.sv`, `src/cdc_fifo_gray_clearable.sv`, `src/multiaddr_decode.sv`, and `src/spill_register_flushable.sv`, so the authors of these @WRoenninger, @fabianschuiki, @zarubaf, @meggiman, and @colluca might want to take a look).
- Assertions that were previously declared as `assert initial` or `assert final` are now immediate assertions in a `initial` or `final` block, respectively.
- `src/rr_arb_tree.sv`, `src/stream_omega_net.sv`, and `src/stream_xbar.sv` previously used a default disable (e.g., `default disable iff (~rst_ni);`), which btw is also not supported by Verilator. Each assertion macro has an `disable iff !rst_ni` instead, so I removed those.

Also, the following files make use of assertions in a pretty advanced/adventurous way (which btw also causes Verilator 5.018 to fail as reported in #229):
- `src/addr_decode_dync.sv`: Includes a huge `always` block with `assume`s all over the place.
- `src/multiaddr_decode.sv`: Again an `always` block with an `assume` in it.

I'd like to ask the respective authors (@WRoenninger and @colluca) to review the changes to these files (you are of course welcome to review changes to the other files as well).

Merging this PR closes #232.

### DIFF
--- a/include/common_cells/assertions.svh
+++ b/include/common_cells/assertions.svh
@@ -6,8 +6,8 @@
 //  - Provides default clk and rst options to simplify code
 //  - Provides boiler plate template for common assertions
 
-`ifndef PRIM_ASSERT_SV
-`define PRIM_ASSERT_SV
+`ifndef COMMON_CELLS_ASSERTIONS_SVH
+`define COMMON_CELLS_ASSERTIONS_SVH
 
 `ifdef UVM
   // report assertion error with UVM if compiled
@@ -34,7 +34,7 @@
 `endif
 
 // Converts an arbitrary block of code into a Verilog string
-`define PRIM_STRINGIFY(__x) `"__x`"
+`define ASSERT_STRINGIFY(__x) `"__x`"
 
 // ASSERT_RPT is available to change the reporting mechanism when an assert fails
 `define ASSERT_RPT(__name)                                                  \
@@ -55,23 +55,23 @@
 
 // Immediate assertion
 // Note that immediate assertions are sensitive to simulation glitches.
-`define ASSERT_I(__name, __prop)           \
-`ifdef INC_ASSERT                          \
-  __name: assert (__prop)                  \
-    else begin                             \
-      `ASSERT_RPT(`PRIM_STRINGIFY(__name)) \
-    end                                    \
+`define ASSERT_I(__name, __prop)             \
+`ifdef INC_ASSERT                            \
+  __name: assert (__prop)                    \
+    else begin                               \
+      `ASSERT_RPT(`ASSERT_STRINGIFY(__name)) \
+    end                                      \
 `endif
 
 // Assertion in initial block. Can be used for things like parameter checking.
-`define ASSERT_INIT(__name, __prop)          \
-`ifdef INC_ASSERT                            \
-  initial begin                              \
-    __name: assert (__prop)                  \
-      else begin                             \
-        `ASSERT_RPT(`PRIM_STRINGIFY(__name)) \
-      end                                    \
-  end                                        \
+`define ASSERT_INIT(__name, __prop)            \
+`ifdef INC_ASSERT                              \
+  initial begin                                \
+    __name: assert (__prop)                    \
+      else begin                               \
+        `ASSERT_RPT(`ASSERT_STRINGIFY(__name)) \
+      end                                      \
+  end                                          \
 `endif
 
 // Assertion in final block. Can be used for things like queues being empty
@@ -82,7 +82,7 @@
   final begin                                                                \
     __name: assert (__prop || $test$plusargs("disable_assert_final_checks")) \
       else begin                                                             \
-        `ASSERT_RPT(`PRIM_STRINGIFY(__name))                                 \
+        `ASSERT_RPT(`ASSERT_STRINGIFY(__name))                               \
       end                                                                    \
   end                                                                        \
 `endif
@@ -93,7 +93,7 @@
 `ifdef INC_ASSERT                                                                        \
   __name: assert property (@(posedge __clk) disable iff ((__rst) !== '0) (__prop))       \
     else begin                                                                           \
-      `ASSERT_RPT(`PRIM_STRINGIFY(__name))                                               \
+      `ASSERT_RPT(`ASSERT_STRINGIFY(__name))                                             \
     end                                                                                  \
 `endif
 // Note: Above we use (__rst !== '0) in the disable iff statements instead of
@@ -106,7 +106,7 @@
 `ifdef INC_ASSERT                                                                              \
   __name: assert property (@(posedge __clk) disable iff ((__rst) !== '0) not (__prop))         \
     else begin                                                                                 \
-      `ASSERT_RPT(`PRIM_STRINGIFY(__name))                                                     \
+      `ASSERT_RPT(`ASSERT_STRINGIFY(__name))                                                   \
     end                                                                                        \
 `endif
 
@@ -157,17 +157,17 @@
 `ifdef INC_ASSERT                                                                        \
   __name: assume property (@(posedge __clk) disable iff ((__rst) !== '0) (__prop))       \
     else begin                                                                           \
-      `ASSERT_RPT(`PRIM_STRINGIFY(__name))                                               \
+      `ASSERT_RPT(`ASSERT_STRINGIFY(__name))                                             \
     end                                                                                  \
 `endif
 
 // Assume an immediate property
-`define ASSUME_I(__name, __prop)           \
-`ifdef INC_ASSERT                          \
-  __name: assume (__prop)                  \
-    else begin                             \
-      `ASSERT_RPT(`PRIM_STRINGIFY(__name)) \
-    end                                    \
+`define ASSUME_I(__name, __prop)             \
+`ifdef INC_ASSERT                            \
+  __name: assume (__prop)                    \
+    else begin                               \
+      `ASSERT_RPT(`ASSERT_STRINGIFY(__name)) \
+    end                                      \
 `endif
 
 //////////////////////////////////
@@ -198,4 +198,4 @@
    `COVER(__name, __prop, __clk, __rst)                                                     \
 `endif
 
-`endif // PRIM_ASSERT_SV
+`endif // COMMON_CELLS_ASSERTIONS_SVH
--- a/include/common_cells/assertions.svh
+++ b/include/common_cells/assertions.svh
@@ -43,12 +43,14 @@
 `define ASSERT_STRINGIFY(__x) `"__x`"
 
 // ASSERT_RPT is available to change the reporting mechanism when an assert fails
-`define ASSERT_RPT(__name)                                                  \
-`ifdef UVM                                                                  \
-  assert_rpt_pkg::assert_rpt($sformatf("[%m] %s (%s:%0d)",                  \
-                             __name, `__FILE__, `__LINE__));                \
-`else                                                                       \
-  $error("[ASSERT FAILED] [%m] %s (%s:%0d)", __name, `__FILE__, `__LINE__); \
+`ifndef ASSERT_RPT
+`define ASSERT_RPT(__name, __desc = "")                                                 \
+`ifdef UVM                                                                              \
+  assert_rpt_pkg::assert_rpt($sformatf("[%m] %s: %s (%s:%0d)",                          \
+                             __name, __desc, `__FILE__, `__LINE__));                    \
+`else                                                                                   \
+  $error("[ASSERT FAILED] [%m] %s: %s (%s:%0d)", __name, __desc, `__FILE__, `__LINE__); \
+`endif
 `endif
 
 ///////////////////////////////////////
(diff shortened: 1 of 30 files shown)


colluca:
@michael-platzer thanks for the PR :)

I will not be able to review this before Monday, but I will definitely have a look then.

michael-platzer:
I reckon that this is a pretty large PR. @niwis Would you rather have me split this into multiple smaller PRs? Perhaps starting with the changes to `assertions.svh` only?

src/stream_omega_net.sv
```diff
@@ -264,51 +267,27 @@ module stream_omega_net #(
...
-          $fatal(1, "Radix %0d is not power of two.", Radix);
-      assert (2**$clog2(NumRouters) == NumRouters) else
-          $fatal(1, "NumRouters %0d is not power of two.", NumRouters);
-      assert ($clog2(NumLanes) % SelW == 0) else
-          $fatal(1, "Bit slicing of the internal selection signal is broken.");
-    end
-    `endif
+    `ASSERT_INIT(radix_not_power_of_2, (2**$clog2(Radix) == Radix) && (Radix > 32'd1), "Radix is not power of two.")
+    `ASSERT_INIT(num_routers_not_power_of_2, 2**$clog2(NumRouters) == NumRouters, "NumRouters is not power of two.")
+    `ASSERT_INIT(bit_sclicing_broken, $clog2(NumLanes) % SelW == 0, "Bit slicing of the internal selection signal is broken.")
```
colluca, line 290: Typo in `bit_sclicing_broken`

colluca (changes requested):
Thanks for the PR @michael-platzer!

It looks good to me for the most part.

One thing I would perhaps change, is to move all `include` statements at the beginning of the files, rather than within the modules. I believe this is more consistent with other include statements and better aligned with one's expectation on where to find included files. Also, for files defining multiple modules e.g. `cdc_2phase_clearable` this would mean we only need to include it once.

Could we also move `COMMON_CELLS_ASSERTS_OFF` to `assertions.svh` for consistency? As of now, some but not all assertions would be disabled, which I think would go against expectation.

Finally, in the PR description (and later in the changelog) I would mention that some assertions which did not define a `disable iff` condition now do after aligning them to use the `ASSERT` macro, and I would list the files affected by this change (e.g. `stream_to_mem.sv`).

src/addr_decode_dync.sv
```diff
@@ -122,21 +124,16 @@ module addr_decode_dync #(
   end
 
   // Assumptions and assertions
-  `ifndef COMMON_CELLS_ASSERTS_OFF
```
phsauter, line 125: This should be left in to turn off only common cells assertions (and for consistency).

src/onehot_to_bin.sv
```diff
@@ -29,10 +31,7 @@ module onehot_to_bin #(
         assign bin[j] = |(tmp_mask & onehot);
     end
 
-`ifndef SYNTHESIS
 `ifndef COMMON_CELLS_ASSERTS_OFF
-    assert final ($onehot0(onehot)) else
-        $fatal(1, "[onehot_to_bin] More than two bit set in the one-hot signal");
-`endif
+    `ASSERT_FINAL(more_than_2_bits, $onehot0(onehot), "[onehot_to_bin] More than two bit set in the one-hot signal")
```
phsauter, line 35: I would remove the [onehot_to_bin] since the final message will already be [ASSERT FAILED] [instance] more_than2_bits ... (file:line)

src/rr_arb_tree.sv
```diff
@@ -170,22 +161,14 @@ module rr_arb_tree #(
...
 
-        `ifndef SYNTHESIS
         `ifndef COMMON_CELLS_ASSERTS_OFF
-          lock: assert property(
-            @(posedge clk_i) disable iff (!rst_ni || flush_i)
-                LockIn |-> req_o && (!gnt_i && !flush_i) |=> idx_o == $past(idx_o)) else
-                $fatal (1, "Lock implies same arbiter decision in next cycle if output is not \
-                            ready.");
+          `ASSERT(lock, LockIn |-> req_o && (!gnt_i && !flush_i) |=> idx_o == $past(idx_o), clk_i, !rst_ni || flush_i, "Lock implies same arbiter decision in next cycle if output is not \
+                            ready.")
```
phsauter, line 166: nitpick: maybe just move the entire message to the next line

src/rr_arb_tree.sv
```diff
@@ -170,22 +161,14 @@ module rr_arb_tree #(
...
           logic [NumIn-1:0] req_tmp;
           assign req_tmp = req_q & req_i;
-          lock_req: assume property(
-            @(posedge clk_i) disable iff (!rst_ni || flush_i)
-                LockIn |-> lock_d |=> req_tmp == req_q) else
-                $fatal (1, "It is disallowed to deassert unserved request signals when LockIn is \
-                            enabled.");
-        `endif
+          `ASSUME(lock_req, LockIn |-> lock_d |=> req_tmp == req_q, clk_i, !rst_ni || flush_i, "It is disallowed to deassert unserved request signals when LockIn is \
+                            enabled.")
```
phsauter, line 171: nitpick: maybe just move the entire message to the next line

src/mem_to_banks_detailed.sv
```diff
@@ -218,18 +220,14 @@ module mem_to_banks_detailed #(
...
-  `ifndef SYNTHESIS
     initial begin
-      assume (DataWidth != 0 && (DataWidth & (DataWidth - 1)) == 0)
-        else $fatal(1, "Data width must be a power of two!");
-      assume (DataWidth % NumBanks == 0)
-        else $fatal(1, "Data width must be evenly divisible over banks!");
-      assume ((DataWidth / NumBanks) % 8 == 0)
-        else $fatal(1, "Data width of each bank must be divisible into 8-bit bytes!");
+      `ASSUME_I(datawidth_not_power_of_2, DataWidth != 0 && (DataWidth & (DataWidth - 1)) == 0,
+               "Data width must be a power of two!")
```
phsauter, line 226: Seems like an insane way to check if it is power of 2. The one used in eg stream_omega_net.sv is a lot saner.

phsauter (changes requested):

src/addr_decode_dync.sv
colluca: What do you think about moving `COMMON_CELLS_ASSERTS_OFF` to `assertions.svh` and have it act globally on all assertions?

phsauter: In principle I am not against it but to me the purpose of COMMON_CELLS_ASSERTS_OFF is to turn off the asserts only in the common cells. Other IPs might still want to use the provided macros though.

But maybe I misunderstood the point of COMMON_CELLS_ASSERTS_OFF.

colluca: What you say makes sense.

Perhaps what we could have in `assertions.svh` is something like the following:
```
`ifndef COMMON_CELLS
`ifdef COMMON_CELLS_ASSERTS_OFF
`undef COMMON_CELLS_ASSERTS_OFF
`endif
`endif
```

Where we could define `COMMON_CELLS` e.g. using Bender, for the `common_cells` source groups only https://github.com/pulp-platform/bender?tab=readme-ov-file#sources.

Technically, users could even achieve the goal you mention without the guard above, by just defining  `COMMON_CELLS_ASSERTS_OFF` on the commands compiling `common_cells` source groups only, but I guess this might not be straightforward to do.

Alternatively, as far as I'm concerned, we could also just cope with a little bit of boilerplate code, and surround all assertions in `common_cells` with the `COMMON_CELLS_ASSERTS_OFF` guard, but I think this PR would be a good moment to do so.

phsauter: I think thats a good approach, maybe we could plan for an automatic define that is added to files of a package, similar to how Bender will automatically add `TARGET_*` defines.  
So something like `PACKAGE_COMMON_CELLS` or `BENDER_COMMON_CELLS`.

include/common_cells/assertions.svh
```diff
@@ -25,24 +25,32 @@
...
-                             __name, `__FILE__, `__LINE__));                \
-`else                                                                       \
-  $error("[ASSERT FAILED] [%m] %s (%s:%0d)", __name, `__FILE__, `__LINE__); \
+`ifndef ASSERT_RPT
+`define ASSERT_RPT(__name, __desc = "")                                        \
+`ifdef UVM                                                                     \
+  assert_rpt_pkg::assert_rpt($sformatf("[%m] %s: %s (%s:%0d)",                 \
+                             __name, __desc, `__FILE__, `__LINE__));           \
+`else                                                                          \
+  $fatal(1, "[ASSERT FAILED] [%m] %s (%s:%0d)", __name, `__FILE__, `__LINE__); \
```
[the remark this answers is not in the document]
michael-platzer: Indeed. Apologies, that change was not meant to make it into the PR :see_no_evil:

src/stream_omega_net.sv
michael-platzer: Thanks for spotting!

src/onehot_to_bin.sv
michael-platzer: Done!

src/rr_arb_tree.sv
michael-platzer: Done and fixed all the line-length lint warnings.

src/rr_arb_tree.sv
michael-platzer: Done and fixed all the line-length lint warnings.

src/mem_to_banks_detailed.sv
michael-platzer: Good point, changed it.

michael-platzer:
Thanks a lot @colluca and @phsauter for the reviews!

> One thing I would perhaps change, is to move all `include` statements at the beginning of the files, rather than within the modules. I believe this is more consistent with other include statements and better aligned with one's expectation on where to find included files. Also, for files defining multiple modules e.g. `cdc_2phase_clearable` this would mean we only need to include it once.

Agree, for the assertion macros this makes more sense, since they are intended to be defined globally anyways.

> Could we also move `COMMON_CELLS_ASSERTS_OFF` to `assertions.svh` for consistency? As of now, some but not all assertions would be disabled, which I think would go against expectation.

We could, but IMHO that is not consistent and would defeat the purpose to disable assertions in the `common_cells` sources only while allowing the assertion macros to be used in other places as well. See my comments above.

> Finally, in the PR description (and later in the changelog) I would mention that some assertions which did not define a `disable iff` condition now do after aligning them to use the `ASSERT` macro, and I would list the files affected by this change (e.g. `stream_to_mem.sv`).

Good point, will add that.

src/addr_decode_dync.sv
colluca: We can keep it as is for now, and eventually address it in a separate PR.

colluca (approved):

michael-platzer:
Please note that `addr_decode_dync.sv` fails to elaborate with the newest Verilator release 5.028 because of issues with the assertion constructs that are unrelated to the changes in this PR. See https://github.com/pulp-platform/common_cells/pull/229#issuecomment-2382908002

phsauter (approved):

niwis (approved):
Thank you very much @michael-platzer! Also @colluca and @phsauter for your thorough reviews! As far as I can tell, all changes are backwards-compatible. In particular, no new `fatal`s were introduced. Could you confirm? In that case, I think this does not require a major release. Please let me know if you have any objections.

michael-platzer:
> As far as I can tell, all changes are backwards-compatible.

To the best of my knowledge, they should be.

> In particular, no new `fatal`s were introduced.

That, I can confirm with certainty. Some `$fatal`s have been converted to `$error`s but there are no new `$fatal`s for sure.

PR #235, opened 2024-09-26
`multiaddr_decode`: Add default port logic

colluca:
In the case of the multi-address decoder, it makes sense to forward a request to the default port if the request's address set is not fully contained in the union of all rules' address sets, i.e. if the intersection is non-empty.

To simplify the module, we expect this union to be externally defined and passed to the module as a parameter.

This PR implements the logic required to implement a default port for the multi-address decoder.


niwis (commented):
Thank you Luca, LGTM in general. Could you elaborate on how the union of all rules should be computed? Maybe there would be a point in adding the corresponding logic in this module (perhaps as a function) to make this easier to reuse.

Since the PR introduces interface changes, I am tagging this as v2. Let me know if you need a release tag, then I can create a release candidate.

PR #236, opened 2024-09-30
stream_join_dynamic: Assert inp readies only for selected streams

ricted98:
Observed behavior: a successful handshake on the output ports asserts `ready=1` on all input ports regardless of `sel_i` selection. This leads to unwanted handshakes on the input channels which have `valid=1` but are not selected. This PR proposes a simple fix to ensure only selected input channels observe `ready=1`.


colluca (approved):
LGTM, thanks @ricted98 :)

niwis (approved):
Thanks @ricted98!

PR #240, opened 2024-10-08
assertions: Keep helper macros defined

michael-platzer:
Another day, another fix for PR #233. This time I realized that undefining a helper macro that is still needed later on is not a smart idea — turns out that due to this the assertion macros were not enabled so far.

Also, a comment in lowRISC's implementation made me realize that the helper macro can be useful outside of the header:

https://github.com/lowRISC/opentitan/blob/28af81ab6e41d9ee6b270f28c7e62761d6c62c31/hw/ip/prim/rtl/prim_assert.sv#L59-L60

### DIFF
--- a/include/common_cells/assertions.svh
+++ b/include/common_cells/assertions.svh
@@ -24,7 +24,7 @@
 // Helper macros //
 ///////////////////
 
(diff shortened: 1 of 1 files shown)


niwis (approved):
Thanks!

PR #242, opened 2024-10-09
Fix inverted reset in delta_counter

olofk:
A recent fix to the reset logic in delta_counter 10dac0ff33 looks like it accidentally inverted the reset logic.


niwis (commented):
Thanks a lot @olofk, great catch! Just one nitpick: could you use the logical negation operator instead? i.e. `!rst_ni`

olofk:
Better now?

FrancescoConti (approved):
LGTM

niwis (approved):
Yep ,thanks a lot!

PR #244, opened 2024-10-17
Fix `fall_through_register` description

mp-17:
The description of the `fall_through_register` says that it does not cut any combinatorial paths. 

This looks wrong, as the `ready` path seems actually cut (`ready_o` depends on `fifo_full` [only](https://github.com/pulp-platform/common_cells/blob/master/src/fall_through_register.sv#L53), which does [not depend on any other interface signal](https://github.com/pulp-platform/common_cells/blob/master/src/fifo_v3.sv#L57)).

### DIFF
--- a/src/fall_through_register.sv
+++ b/src/fall_through_register.sv
@@ -9,9 +9,10 @@
 // specific language governing permissions and limitations under the License.
 
 // Fall-through register with a simple stream-like ready/valid handshake.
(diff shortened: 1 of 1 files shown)


niwis (changes requested):
Could we change this to "does not cut the combinatorial path on the valid and data signals" or similar? To me, this seems more relevant

mp-17:
Done, I specified both!

niwis (approved):

PR #247, opened 2024-11-05
treewide: Replace non-resettable FFs

colluca:
Following https://github.com/pulp-platform/snitch_cluster/pull/154, there are still some non-resettable FFs in the Snitch cluster, that can be traced back to the `isochronous_spill_register` IP.

This PR is a placeholder for addressing this issue. The current solution has to be revised to ensure safety against metastability conditions.


niwis:
@colluca any updates on this?

colluca:
Hi @niwis, haven't had the time to look into this. I'm also not an expert in CDCs, reset synchronization, etc. so if anyone could provide input on this it would be useful, as properly handling the resets of the two domains is not trivial.

niwis (approved):

PR #249, opened 2024-11-22
[Verilator] Don't import deprecated SRAM with verilator

Scheremo:
The deprecated SRAM imported with `simulation` breaks with verilator. This PR excludes said module from being imported with `verilator`.

### DIFF
--- a/Bender.yml
+++ b/Bender.yml
@@ -113,7 +113,7 @@ sources:
       # Level 4
       - src/mem_to_banks.sv
 
-  - target: simulation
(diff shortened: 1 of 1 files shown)


niwis (approved):
LGTM. Not sure if `simulation` should be set together with `verilator` in the first place, though (see [here](https://github.com/pulp-platform/bender/blob/46ade8e6b97b8d27b601ed7348f541c399366c25/src/cmd/script.rs#L222), although this requirement might be pre-Verilator5).

PR #254, opened 2025-03-17
Fix mismatch in id_queue triggering obscure errors in QuestaSim

FrancescoConti:
The error has been reported in the past in a closed issue in PULPissimo:
https://github.com/pulp-platform/pulpissimo/issues/71#issuecomment-522698136

but is also triggered in other conditions (e.g., on work that is not yet open-source).
The fix is trivial.


niwis (commented):
actually, should this also be fixed here? https://github.com/pulp-platform/common_cells/blob/9540f880fa39a599e157385c87bbafab915951b0/src/id_queue.sv#L264

PR #258, opened 2025-06-20
id_queue: Simplify existence lookup for Verilator speed-up

MaxWipfli:
This `always_comb` block previously consumed 9% of total simulation runtime for Cheshire on Verilator (pulp-platform/cheshire#230). With this change, this is reduced to a vanishing amount (reported as 0.00%).


niwis (changes requested):
Thank you @MaxWipfli, this is a great simplification! Just one request: could you consistently use the bit operators on the RHS? (`~` instead of `!`, `&` instead of `&&`).

niwis (approved):

PR #262, opened 2025-07-18
Fix SpyGlass CDC Lint issues in CDC FIFOs

Scheremo:
This PR fixes CDC linting issues of type `Ac_cdc01` in the modules `cdc_fifo_gray` and `cdc_fifo_gray_clearable`.
Currently, the tool flags "ClkCross Data Hold" violations of some (but not all) paths originating from single bits of `wptr_q`.

The issue arise from the formulation of pushing new data to the FIFO. Currently, the FIFO push is enabled by the expression 
`src_valid_i & src_ready_o & (wptr_bin[LOG_DEPTH-1:0] == i)`, which in SpyGlass CDC is viewed as the qualifier of the crossing. Unfortunately this leads SpyGlass CDC to evaluate individual bits of `wptr_q` as being qualifiers for multiple clock domain crossings, which leads to the pointer not being properly recognized as a CDC FIFO write pointer. In turn, the tool will complain that (some) of those `wptr_q` bits violate fast-to-slow clock domain data hold checks, since they are allowed to change multiple times within a target domain clock cycle.

Reformulating (practically) the same functionality by removing `wptr` from the enable condition fixes this, enabling this code to pass strict_qual for SpyGlass CDC linting.

### DIFF
--- a/src/cdc_fifo_gray.sv
+++ b/src/cdc_fifo_gray.sv
@@ -192,10 +192,12 @@ module cdc_fifo_gray_src #(
 
   // Data FIFO.
   assign async_data_o = data_q;
-  for (genvar i = 0; i < 2**LOG_DEPTH; i++) begin : gen_word
-    `FFLNR(data_q[i], src_data_i,
-          src_valid_i & src_ready_o & (wptr_bin[LOG_DEPTH-1:0] == i), src_clk_i)
+
+  always_comb begin
+    data_d                          = data_q;
+    data_d[wptr_bin[LOG_DEPTH-1:0]] = src_data_i;
   end
+  `FFLARN(data_q, data_d, src_valid_i & src_ready_o, '0, src_clk_i, src_rst_ni)
 
   // Read pointer.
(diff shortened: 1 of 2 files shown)


phsauter:
Both changes look good to me. 

Note that the FF type of the data FIFO in the source-side has also been changed from no-reset to one with reset. This should not cause a functional difference and reset related requirements and guarantees are unaffected. Warm reset in cdc_fifo_gray is marked as being dangerous and in the clearable version it is explicitly handled by the reset controller.

Scheremo:
> Both changes look good to me.
> 
> Note that the FF type of the data FIFO in the source-side has also been changed from no-reset to one with reset. This should not cause a functional difference and reset related requirements and guarantees are unaffected. Warm reset in cdc_fifo_gray is marked as being dangerous and in the clearable version it is explicitly handled by the reset controller.

Thanks for the review! I swapped the `FFLNR` with `FFLARN` because I am a reset fan. If this change is an issue we can also revert it, SpyGlass does not complain about the FIFO not being resettable; this was just me freewheeling.

PR #275, opened 2025-12-15
Add `iomsb`, `min` and `max` functions

colluca:
I believe this is a more convenient place to have the `iomsb` function, rather than every repo redefining it. Also adding the `min` function used here https://github.com/pulp-platform/cluster_icache/pull/32.

### DIFF
--- a/src/cf_math_pkg.sv
+++ b/src/cf_math_pkg.sv
@@ -64,4 +64,9 @@ package cf_math_pkg;
         return (value != 0) && (value & (value - 1)) == 0;
     endfunction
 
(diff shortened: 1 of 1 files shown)


phsauter:
I think this is a good addition.  
In principle a second declaration imported via `pkg::*` should just hide the first one and shouldn't produce an error but I think we should actually check this first in the major tools (yosys-slang, VCS, Verilator, Questasim, Synopsys DC) to make sure the tools do not produce errors.  
If they do produce errors, we would also have to add this as part of a v2.

colluca:
Fixed the `min` function (which was actually calculating the maximum) and added the `max` function.

PR #283, opened 2026-03-18
[HW] Fixed stream fork to correctly propagate ready signal

Camma2001:
Modifications necessary to ensure the correct functionality of the ready signal when stream_fork instances are in cascade.


phsauter:
@Camma2001 can you explain this further, otherwise I will close it for now.

dcammarata01:
I need to check more thoroughly whether these modifications are actually needed, so it is probably better to close it now.

PR #290, opened 2026-04-21
treewide: Namespace all modules with a `cc_` prefix 

colluca:
- Namespace all modules with `cc_` prefix
- Rename `stream_test` package to `cc_test_pkg`


phsauter:
> **TODOs**
> 
>     * [ ]  Should we prefix also all macros?
> 
>     * [ ]  Is `simulate.sh` a legacy script or should it be updated/added to the CI?

No lets leave the macros,

Probably legacy but I am not sure, we need to remove deprecated things later anyway, lets wait for that.

PR #291, opened 2026-04-22
pkg: Merge all packages into a single `cc_pkg`

colluca:
Merges all IP-specific packages into one `common_cells_pkg.sv` file.
My proposal is to name the package `cc` for more concise reference, e.g. `cc:iomsb` vs. the previous `cf_math_pkg::iomsb`.


phsauter:
Some linters complain when there are:
1. Packages not ending in eg `_pkg`
2. Files with a different name than their module/package

So I would propose naming it `cc_pkg` instead. It is still relatively concise but should be better for linting.

colluca:
Sounds like a reasonable compromise. I just pushed the changes renaming the package and file to `cc_pkg`.

Issue #256, opened 2025-06-04
Missing description for `mv_filter` IP


colluca:
https://github.com/pulp-platform/common_cells/blob/47b020ab3355116f750a5f14c67a409d5c11c5df/README.md?plain=1#L65

phsauter:
Solved with #293

PR #293, opened 2026-04-23
Reimplement `serial_deglitch` and `mv_filter`

phsauter:
Defines a new expected behavior of `serial_deglitch` and `mv_filter`.  
The second is different enough that I decided to change its name to `majority_vote_filter`.

Replaces #252, addresses #256


src/majority_vote_filter.sv
```diff
@@ -0,0 +1,57 @@
...
+    input  logic en_i,
+    input  logic d_i,
+    output logic q_o
+);
+    localparam int unsigned CNT_WIDTH = $clog2(WINDOW_LEN+1);
+
+    logic [CNT_WIDTH-1:0] popcount_o;
+    logic [WINDOW_LEN-1:0] history_d, history_q;
+
+    assign history_d = {history_q[WINDOW_LEN-2:0], d_i};
```
colluca, line 40: If we allow `WINDOW_LEN = 1` this part-select underflows. I don't see any use case for `WINDOW_LEN = 1` but I might be mistaken. In any case, we should either forbid this through an assertion, or somehow fix this corner case.

src/majority_vote_filter.sv
```diff
@@ -0,0 +1,57 @@
...
+        .data_i(history_q),
+        .popcount_o(popcount_o)
+    );
+
+    assign q_o = (popcount_o >= CNT_WIDTH'(THRESHOLD));
+
+`ifndef COMMON_CELLS_ASSERTS_OFF
+    `ASSERT_INIT(WindowLenGtZero, WINDOW_LEN >= 1, "majority_vote_filter: WINDOW_LEN must be >= 1")
+    `ASSERT_INIT(ThresholdGtZero, THRESHOLD >= 1,  "majority_vote_filter: THRESHOLD must be >= 1")
+`endif
```
colluca, line 59: We might want to add an upper-bound assertion on `THRESHOLD` as well, e.g.:
```
`ASSERT_INIT(ThresholdLtWindow, THRESHOLD <= WINDOW_LEN, "majority_vote_filter: THRESHOLD must be <= WINDOW_LEN")
```

colluca (changes requested):
Mostly looks good to me, there's only one corner case that could possibly create issues.

src/majority_vote_filter.sv
phsauter: Good catch, I changed it

src/majority_vote_filter.sv
phsauter: Makes sense, added

colluca (approved):

colluca:
@phsauter I noticed this somehow went missing on v2-dev. Any idea how? Just mentioning cause we should push these commits again.

PR #298, opened 2026-05-02
Restore serial deglitch and majority vote filter

phsauter:
Restoring the changes from #293 into the v2-dev branch.  
I think I rebased on a stale version of v2-dev at some point and then pushed into v2-dev to get rid of the difference to main.  
Clearly I was not careful enough.


phsauter:
@colluca This should fix it, sorry for not being more careful.

colluca (approved):
Np, can happen ofc. LGTM :)

PR #304, opened 2026-05-22
Refactor cdc_2phase_clearable and extend its tests

phsauter:
- Add formal verification for the CDC reset controller halfs
- Add formal verification for the CDC reset controller top
- Expand the previous `cdc_2phase_clearable` timed testbench and add asserts to check various conditions
- Refactor the `cdc_2phase_clearable` using the new tests to check correctness, the goal was to cleanly separate it into a source and destination side and only have the async wires going between them in the top-level.
- Fixed parameters that were not passed down properly

This is a big one, we can gladly discuss it but if you want to review something focus on the `cdc_2phase_clearable` itself and if the parameters do what you would expect. 
I am very confident this does _not_ change the reset/clear behavior.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,6 +9,8 @@ build
 formal/fifo
 formal/counter
 formal/fall_through_register
+formal/cdc_reset_ctrlr_half
+formal/cdc_reset_ctrlr_composed
 *.check
 *.vcd
 obj_dir/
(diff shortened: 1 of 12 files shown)


colluca (commented):
Yes, this is indeed very tricky but it does look good to me at a high level.

Can we include the formal tests in the CI, if they are not already?

PR #306, opened 2026-05-28
Add Verilator elaboration to CI

colluca:
- Remove legacy Travis CI and Make targets used only therein
- Add per-module Verilator elaboration target to Makefile
- Fix Verilator elaboration errors
  - Implicit casting to enum (actual error that I got in Snitch, which triggered the need for this PR)
  - Errors with the default parameters
- Add elaboration tests to both Github and Gitlab CIs
- Add a script for environment setup on IIS machines (used also by Gitlab CI)


src/cc_stream_arbiter.sv
```diff
@@ -15,7 +15,7 @@
 
 module cc_stream_arbiter #(
     parameter type      DATA_T = logic,   // Vivado requires a default value for type parameters.
-    parameter integer   N_INP = -1,       // Synopsys DC requires a default value for parameters.
+    parameter integer   N_INP = 0,        // Synopsys DC requires a default value for parameters.
```
phsauter, line 18: Since the address decoders now get a valid struct as defaults, should we also have a valid default here?  
So it would be `N_INP=1` instead of zero which below still leads to a very weird `logic [-1:0]`

Also this should be a `int unsigned` which is a uint32, integer is a 4-state datatype which does not really make sense here.

src/cc_stream_arbiter_flushable.sv
```diff
@@ -15,7 +15,7 @@
 
 module cc_stream_arbiter_flushable #(
     parameter type      DATA_T = logic,   // Vivado requires a default value for type parameters.
-    parameter integer   N_INP = -1,       // Synopsys DC requires a default value for parameters.
+    parameter integer   N_INP = 0,        // Synopsys DC requires a default value for parameters.
```
phsauter, line 18: Same as in `cc_stream_arbiter_flushable`

phsauter (changes requested):
Minor changes that might be nice to have but are not strictly required.

Makefile
```diff
@@ -1,9 +1,22 @@
-# SPDX-License-Identifier: Apache-2.0
-
+BENDER    ?= bender
 VERILATOR ?= verilator
 
-all: cc_ecc_encode cc_ecc_decode
```
phsauter, line 5: A default would be nice, eg:
```
.PHONY: all
all: elab
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  elaborate:
+    name: Elaborate all RTL modules w/ Verilator
+    runs-on: ubuntu-latest
+    container:
+      image: hpretl/iic-osic-tools:2026.04
+      options: --user root
+    steps:
+      - uses: actions/checkout@v4
```
phsauter: Use `@v5` or `@v6` so we can leave it for a while.

phsauter (changes requested):
Some Ci fixes

src/cc_stream_arbiter.sv
colluca: Indeed Spyglass doesn't like that default value, so it is changed in a separate PR (https://github.com/pulp-platform/common_cells/pull/308/changes#diff-9c58aee2e6d49045f3b90aed1154b8f6f162c53cfcbd8c303c33a524c42096edR18).

Regarding `integer` that's a good point. Probably better to address it globally in a separate PR (opened an issue in the meantime https://github.com/pulp-platform/common_cells/issues/309).

Makefile
colluca: If I'm not mistaken, `all` defaults to the first target in a Makefile. Might still be better to make it explicit as you suggest.

colluca: Sorry, `all` is not defined by default. But the first target in a Makefile is the default goal ([reference](https://www.gnu.org/software/make/manual/html_node/Goals.html)):
> By default, the goal is the first target in the makefile (not counting targets that start with a period). Therefore, makefiles are usually written so that the first target is for compiling the entire program or programs they describe. If the first rule in the makefile has several targets, only the first target in the rule becomes the default goal, not the whole list. You can manage the selection of the default goal from within your makefile using the .DEFAULT_GOAL variable (see [Other Special Variables](https://www.gnu.org/software/make/manual/html_node/Special-Variables.html)).

src/cc_stream_arbiter_flushable.sv
colluca: Same as above

Makefile
phsauter: Ah okay, thats fine then.

PR #311, opened 2026-05-31
Cleanup module port and parameter lists

colluca:
Only stylistic changes in this one, mostly aligned with the [LowRisc style guide](https://github.com/lowRISC/style-guides/blob/master/VerilogCodingStyle.md).

- Add `_i`/`_o` suffixes to all module ports
- Convert all "modifiable" or "derived" module parameters from ALL_CAPS to UpperCamelCase. Type parameters are excluded, and addressed in the next point. This point includes:
  - Module parameter port lists (parameters and dependent localparams)
  - Module body localparams derived from module parameters
  - All instantiation sites using named parameter binding
  - Comments and string literals that referenced old names
- Align all type parameter names to the *_t convention (e.g. `T`, `dtype`, `type_t`, `DataType`, `DATA_T` are all converted to `data_t`)
- Replace all string parameters with enumerated types, defined in `cc_pkg.sv`

**TODO**
- [x] merge after https://github.com/pulp-platform/common_cells/pull/310


phsauter (approved):
This took a while to get through but it looks good, I don't think any bug was introduced or something was missed.

PR #312, opened 2026-06-03
Extend Spyglass lint to lint/lint_rtl goal

colluca:
The Spyglass lint would previously only run the `DesignRead` goal.
This PR extends it to the `lint/lint_rtl` goal (which includes the `DesignRead` goal).

TODO:
- [x] merge after https://github.com/pulp-platform/common_cells/pull/311


phsauter:
I will have to think about whether I like the blocking assignments in `clk_int_div`, as I understand it they are only needed for simulation purposes so we might want to have explicit synthesis and simulation versions there.

PR #314, opened 2026-06-04
treewide: Make FFs clearable

niwis:
Add an active-high synchronous clear `clr_i` to all modules, clearing all FFS to their reset value (analogous to `rst_ni`). Streamline the codebase by consistently using register macros for sequential logic.

Design principles:
- `clr_i` shall clear *all* FFs, similar to `rst_ni`.
- Every sequential module shall have a `clr_i`.
- Consequently, if a module already has a functional clear that clears *all* FFs, rename that port to `clr_i`. If a an existing functional clear only clears a subset of FFs, rename that port to something distinctively else and add `clr_i`.

The previous `stream_arbiter` wrapper is now redundant and replaced by the previous `stream_arbiter_flushable`.

Sequential modules where *no* synchronous clear was added:
- `cdc_*`
- `clk_*`
- `edge_propagator*`
- `isochronous_*`
- `rstgen*`


src/cc_counter.sv
```diff
@@ -17,7 +17,7 @@ module cc_counter #(
 )(
     input  logic             clk_i,
     input  logic             rst_ni,
-    input  logic             clear_i, // synchronous clear
+    input  logic             clr_i,   // synchronous clear
```
phsauter: While we are at it, could you make sure the comments here are consistent? Here it is lower case, in cb_filter it is uppercase. I think all modules should have a comment for each clock, the reset and the synchronous clear.  
Active high is assumed but not stated, the resets active low is stated, where applicable (eg credit_counter) an additional bit of comment after a semicolon is useful.

src/cc_fifo.sv
```diff
@@ -22,6 +23,7 @@ module cc_fifo #(
 )(
     input  logic  clk_i,            // Clock
     input  logic  rst_ni,           // Asynchronous reset active low
+    input  logic  clr_i,            // synchronous clear
```
phsauter: The difference between `clr_i` and `flush_i` (the former clears the elements in the fifo to zero as well as clearing the pointers, which might have a slight energy penalty) should be stated since it is hard to understand without looking at the code.

src/cc_stream_fifo.sv
```diff
@@ -23,6 +23,7 @@ module cc_stream_fifo #(
 ) (
     input  logic                 clk_i,      // Clock
     input  logic                 rst_ni,     // Asynchronous reset active low
+    input  logic                 clr_i,      // Synchronous clear
```
phsauter: Again state difference between clr and flush in comment.

src/cc_stream_fifo_optimal_wrap.sv
```diff
@@ -19,6 +19,7 @@ module cc_stream_fifo_optimal_wrap #(
 ) (
     input  logic                 clk_i,      // Clock
     input  logic                 rst_ni,     // Asynchronous reset active low
+    input  logic                 clr_i,      // Synchronous clear
```
phsauter: Again state difference between clr and flush in comment.

src/cc_spill_register_flushable.sv
```diff
@@ -21,6 +22,7 @@ module cc_spill_register_flushable #(
 ) (
   input  logic  clk_i   ,
   input  logic  rst_ni  ,
+  input  logic  clr_i   ,
```
phsauter: Again state difference between clr and flush in comment.

phsauter (changes requested):
Some cleanup to explain the difference between clr and other signals and to unify the comments a bit more. Should be easily doable by LLM.

Additionally the Readme needs updating:
- cc_stream_arbiter_flushable is removed
- rst_ni and clr_i need to be mentioned and their purpose explained
- The use of the FF macros should be mentioned in the Readme.

niwis:
Thanks, should all be addressed now.

phsauter (approved):
LGTM!

PR #315, opened 2026-06-05
pkg: use unsigned int over integer

phsauter:
- Use `int` (2 state) instead of `integer` (3 state)
- Use unsigned where possible
- Use if/else to distinguish from hardware MUX
- Avoid implicit variables from function name

Has some minor functional differences but only for values that were previously guarded against. Otherwise it should be functionally identical.

### DIFF
--- a/src/cc_pkg.sv
+++ b/src/cc_pkg.sv
@@ -15,47 +15,59 @@ package cc_pkg;
   /// Ceiled Division of Two Natural Numbers
   ///
   /// Returns the quotient of two natural numbers, rounded towards plus infinity.
-  function automatic integer ceil_div (input longint dividend, input longint divisor);
-    automatic longint remainder;
+  function automatic longint unsigned ceil_div (
+    input longint unsigned dividend,
+    input longint unsigned divisor
+  );
+    automatic longint unsigned quotient = 0;
 
     `ifndef SYNTHESIS
     `ifndef COMMON_CELLS_ASSERTS_OFF
-    if (dividend < 0) begin
-      $fatal(1, "Dividend %0d is not a natural number!", dividend);
-    end
-
-    if (divisor < 0) begin
-      $fatal(1, "Divisor %0d is not a natural number!", divisor);
-    end
-
     if (divisor == 0) begin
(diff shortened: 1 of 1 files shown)


src/cc_pkg.sv
```diff
@@ -15,47 +15,62 @@ package cc_pkg;
...
     `endif
 
     remainder = dividend;
-    for (ceil_div = 0; remainder > 0; ceil_div++) begin
+    for (; remainder >= divisor; quotient++) begin
       remainder = remainder - divisor;
     end
+    if (remainder != 0) begin
+      quotient++;
+    end
```
colluca, line 35: Maybe we could even remove the loop altogether to speedup elaboration, not sure why it was done with a loop:
```
return (dividend + divisor - 1) / divisor;
```

src/cc_pkg.sv
```diff
@@ -15,47 +15,62 @@ package cc_pkg;
...
-  function automatic bit is_power_of_2 (input integer unsigned value);
-    return (value != 0) && (value & (value - 1)) == 0;
+  function automatic bit is_power_of_2 (input int unsigned value);
+    automatic int unsigned value_without_lsb;
+
+    if (value == 32'd0) begin
+      return 1'b0;
+    end
+
+    value_without_lsb = value & (value - 32'd1);
```
colluca: I think the name `value_without_lsb` is a bit misleading, at least if reading LSB as "least significant bit".

If I'm not mistaken, this calculates `value_without_lowest_set_bit`.

colluca (commented):
Looks correct to me. Just a few nitpicks in the comments.

src/cc_pkg.sv
phsauter: The problem with that is that it can overflow. If we limit the input to 32bit we can do it though.  
Did you ever see one where an operand exceeds 4 billion? I would assume no.

phsauter: I used another solution without loop.

src/cc_pkg.sv
phsauter: Good point, changed it.

PR #316, opened 2026-06-17
ci: Add Slang pulp-action for linting

phsauter:
Updates the GitHub RTL checks to be more useful for future PRs.

I originally tried running the checks in the OSEDA/iic-osic-tools container and added the same free-space setup we use in croc, because the later CDC verification work
needs `sby`. It worked but the image is large and makes the public GitHub CI slower and more sensitive to runner disk space. Since the heavier Verilator-style flow
is already covered in GitLab CI, I switched the GitHub side to a light setup.

  The current version keeps two checks in:
  - `Slang Elaboration` through `pulp-platform/pulp-actions/slang`, with reviewdog annotations
  - `Verible Lint`, using a pinned static Verible release

I also fixed the lint findings that showed up with fatal Verible instead of making lint non-fatal. That includes the `CbEgSeeds` rename, moving a few waivers next to
the code they apply to, and some small style cleanups like line wrapping and whitespace.


.github/workflows/lint.yml
```diff
@@ -1,20 +0,0 @@
...
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v6
-      - uses: chipsalliance/verible-linter-action@main
-        with:
-          paths: |
-            ./src
-          extra_args: "--waiver_files lint/common_cells.style.waiver --rules=-interface-name-style --lint_fatal"
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          reviewdog_reporter: github-check
```
micprog, line 20: The reviewdog reporting is a nice way to visualize the lint errors in a PR, which I don't think the new solution supports.

micprog (commented):
I remember the oseda container being quite large, which takes a while to properly download, I personally try to stick to default runners unless the larger tools are required, but overall this is still a workable solution. One small item, don't know how you want to handle.

.github/actions/free-space/action.yml
colluca, line 1: In Snitch I use this https://github.com/jlumbroso/free-disk-space. Might be better than maintaining our own custom action.

.github/scripts/run_verible_lint.sh
```diff
@@ -0,0 +1,15 @@
...
+
+set -euo pipefail
+
+command -v verible-verilog-lint
+
+find src -type f \( -name '*.sv' -o -name '*.svh' -o -name '*.v' \) -print0 \
+  | xargs -0 verible-verilog-lint \
+      --waiver_files lint/common_cells.style.waiver \
+      --rules=-interface-name-style \
+      --lint_fatal=false
```
colluca: Why is `lint_fatal` set to false now?

colluca (commented):
Looks good to me, just have a comment and a question. I don't find the reviewdog reporting particularly useful personally, so from that point of view no difference for me.

.github/workflows/lint.yml
phsauter: It is large yes, we are planing to make a reduced CI image in the future for this reason.  
The problem is that now I am using Yosys SBY for the formal verification and I think it makes sense to have this run in the public facing CI as well, not just internally.

.github/scripts/run_verible_lint.sh
phsauter: Resolved by fixing the lints so it can be fatal (reviewdog internally also used it as non-fatal but now it is).

phsauter:
Sorry for being that guy but I had a change of heart, it now uses the Slang action made by Nils.

.github/workflows/ci.yml
```diff
@@ -6,20 +6,59 @@ name: ci
...
     steps:
-      - uses: actions/checkout@v6
-      - name: Run Verilator elaboration check
-        # Use a login shell to ensure that the environment is set up correctly
-        shell: bash -l -e -o pipefail {0}
+      - name: Checkout repository
+        uses: actions/checkout@v6
+      - name: Install Verible
         run: |
-          make vlt-elab
```
colluca, line 25: If I understand correctly, the motivation for removing this is that the oseda container is too larger for some runners?

I think this is fine, since we anyways have it in the Gitlab CI, but I would update the PR description accordingly.

src/cc_pkg.sv
```diff
@@ -96,7 +96,7 @@ package cc_pkg;
   } cb_seed_t;
 
   // example seeding struct
-  localparam cb_seed_t [2:0] cb_eg_seeds = '{
+  localparam cb_seed_t [2:0] CbEgSeeds = '{
```
colluca, line 99: Would also update in PR description

colluca (approved):
LGTM

.github/workflows/ci.yml
phsauter: Yes exactly, I think its okay-ish that the Gitlab CI handles it.

src/cc_pkg.sv
phsauter: Done

PR #317, opened 2026-06-17
Add formal and timed verification for cc_cdc_2phase_clearable

phsauter:
I split #304 into two easier to handle PRs and had them review by Opus 4.8 and GPT-5.5 which discovered some minor uncovered cases in the formal verification.

- Fixed parameters that were not passed down properly
- Add formal verification for the CDC reset controller halfs
- Add formal verification for the CDC reset controller top
- Expand the previous `cdc_2phase_clearable` timed testbench and add asserts to check various conditions
- Add the new testbench to the simulation CI action and creates a new formal verification action


formal/Makefile
```diff
@@ -1,21 +1,28 @@
...
+# - Robert Balas <balasr@iis.ee.ethz.ch>
+# - Philippe Sauter <phsauter@iis.ee.ethz.ch>
+#
+# Description: Formal Verification Targets
+# Dispatch SymbiYosys proof runs for common_cells formal properties, including
+# reset-controller half and composed CDC clear-sequencing checks.
 
 YOSYS	?= yosys
-SBY	?= sby
+SBY	?= oseda sby
```
colluca: I would just keep the executable name (I suppose `sby`) here for people who have it natively installed, and move the override (`oseda sby`) to the `iis-env.sh` script.

formal/run_cdc_reset_ctrlr_composed.sh
```diff
@@ -0,0 +1,25 @@
...
+#
+# Authors:
+# - Philippe Sauter <phsauter@iis.ee.ethz.ch>
+#
+# Description: Composed Reset-Controller Proof Runner
+# Run the composed reset-controller SymbiYosys proof from the formal directory.
+
+set -euo pipefail
+
+: "${SBY:=oseda sby}"
```
colluca: Same as above, if possible.

formal/run_cdc_reset_ctrlr_half.sh
```diff
@@ -0,0 +1,27 @@
...
+# - Philippe Sauter <phsauter@iis.ee.ethz.ch>
+#
+# Description: Reset-Controller Half Proof Runner
+# Run the reset-controller half SymbiYosys proof from the formal directory.
+
+set -euo pipefail
+
+script_dir="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+
+: "${SBY:=oseda sby}"
```
colluca: Same as above.

.gitignore
```diff
@@ -9,6 +9,8 @@ build
 formal/fifo
 formal/counter
 formal/fall_through_register
+formal/cdc_reset_ctrlr_half
+formal/cdc_reset_ctrlr_composed
```
colluca: Can the artifacts of the formal verification be generated in a build directory? So that we can simply gitignore that and don't have to list each and every module here.

.gitlab-ci.yml
```diff
@@ -82,3 +88,7 @@ tests:
       #   PARAM1: [-GDEPTH=1, -GDEPTH=2, -GDEPTH=3, -GDEPTH=4, -GDEPTH=5]
       #   PARAM2: [-GGRAY=0, -GGRAY=1]
 
+formal:
+  timeout: 30min
+  script:
+    - make -B -C formal cdc_reset_ctrlr_half.check cdc_reset_ctrlr_composed.check
```
colluca, line 95: Just out of curiosity, do the other goals fail? i.e. is there a reason not to have `make all`?

colluca (changes requested):
Looks good to me. Just have a few nitpicks and questions.

formal/Makefile
phsauter: Done

formal/run_cdc_reset_ctrlr_composed.sh
phsauter: handled by the makefile now

formal/run_cdc_reset_ctrlr_half.sh
phsauter: handled by the makefile now

.gitignore
phsauter: Yes good point, moved to `formal/build`

.gitlab-ci.yml
phsauter: Yes they fail, the verification in this repo is/was not in a good state.  
I wanted to do the CDC stuff first then I will go over the remaining formal checks and adjust them and then I will add it back.

colluca (approved):

Issue #289, opened 2026-04-17
Move `sync` to tech_cells


phsauter:
Often technologies have dedicated synchronization cells, therefore the `sync` cells should be part of technology specifics and be in `tech_cells` instead. We can have a generic FF implementation here that uses the tech_cells version via a define or a similar system to make it so this repo does not depend on tech_cells.

phsauter:
Closed with #319

PR #319, opened 2026-06-25
Move `sync` to tech_cells_generic

colluca:
Addresses https://github.com/pulp-platform/common_cells/issues/289.

I completely removed `cc_sync` within common_cells. I think this should be fine since tech_cells is already a dependency of common_cells.

**TODO**
- [x] Merge https://github.com/pulp-platform/tech_cells_generic/pull/43


phsauter:
I had to change the previous formal verification setup to make this work now that some of the files (tc_sync) are handled by bender instead of being directly available.

Issue #69, opened 2020-02-21
Truncation of `usage_o` port in `fifo_v3` when parameter `DEPTH` is a power of 2


WRoenninger:
The `usage_o` port of `fifo_v3` has currently a width of `$clog(DEPTH)`.
This causes the pointer to wrap around to all zeros, if the FIFO has a power of 2 `DEPTH` and is full. This is an issue for the following exapmle:
```systemverilog
localparam int unsigned Depth     = 32'd16;
localparam logic [3:0]  Threshold = 4'hF;
logic [3:0] usage;
if (usage > threshold) begin
  /* do something */
end
```
This `if` statement would not trigger in this case.
This could be resolved by assigning the whole `status_cnt_q` to the `usage_o` port instead of the truncated one as is currently the case.

The workaround I am using currently is prepending the `full_o` signal to the `ussage_o`, however this causes jumps in this new usage pointer value when the FIFO is not configured to a power of 2 `DEPTH`.

zarubaf:
Good catch and thanks for reporting. We should probably fix this in a `fifo_v4` or a breaking change (major version bump) in the common cells repo (which is unfortunately anyway required near term).

PR #321, opened 2026-06-25
fifo: Fix truncation of usage_o for power-of-2 depth (#69)

colluca:
Addresses https://github.com/pulp-platform/common_cells/issues/69.

When Depth is a power of 2, `status_cnt_q` needs `AddrDepth+1` bits to represent a full FIFO (e.g. Depth=16 -> `status_cnt_q=5'b10000`). The previous `usage_o` port was `[AddrDepth-1:0]`, so the MSB was silently dropped and a full FIFO appeared empty (usage==0).

Fix by widening usage_o to `[AddrDepth:0]` and assigning the full `status_cnt_q`. The same fix is applied to the cc_stream_fifo wrapper.

### DIFF
--- a/src/cc_fifo.sv
+++ b/src/cc_fifo.sv
@@ -26,7 +26,7 @@ module cc_fifo #(
     // status flags
     output logic  full_o,           // queue is full
     output logic  empty_o,          // queue is empty
-    output logic  [AddrDepth-1:0] usage_o,  // fill pointer
+    output logic  [AddrDepth:0] usage_o,  // fill pointer
     // as long as the queue is not full we can push new data
     input  data_t data_i,           // data to push into the queue
     input  logic  push_i,           // data is valid and can be pushed to the queue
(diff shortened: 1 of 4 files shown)


phsauter:
I think it might be cleaner to split this into two and have this in the port params:
```
localparam int unsigned UsageWidth = cc_pkg::idx_width(Depth + 1);
```

And then this afterwards for the pointers:
```
localparam int unsigned PtrWidth   = cc_pkg::idx_width(Depth);
```

So everything still has the 'correct' minimum widths required.

Edit: We could even consider adding a function like `cc_pkg::cnt_width(MaxCnt)` that calculates the width required to represent the count/occupancy, meaning all count values from 0 to MaxCnt, inclusive.

colluca:
> I think it might be cleaner to split this into two and have this in the port params:
> 
> ```
> localparam int unsigned UsageWidth = cc_pkg::idx_width(Depth + 1);
> ```
> 
> And then this afterwards for the pointers:
> 
> ```
> localparam int unsigned PtrWidth   = cc_pkg::idx_width(Depth);
> ```
> 
> So everything still has the 'correct' minimum widths required.
> 
> Edit: We could even consider adding a function like `cc_pkg::cnt_width(MaxCnt)` that calculates the width required to represent the count/occupancy, meaning all count values from 0 to MaxCnt, inclusive.

Agree with everything, should be fixed now.

PR #324, opened 2026-06-29
Add legacy wrappers for new `cc_` modules and package

ricted98:


colluca (approved):
Thanks for the contribution, I can later take care of rebasing this on v2-dev.

I had manual look at the changes and couldn't spot anything wrong. Let me know when it works in CVA6/Cheshire and I'll merge it.

ricted98:
Cheshire CI passes. I think they are good to go. One point I noted would be to also add the legacy packages in the `test` folder. Do we want to do it in this PR?

colluca:
That makes sense yes. If the changes are already in v2-test we might as well do it already.

Issue #323, opened 2026-06-25
id_queue: `empty` used without an explicit declaration (implicit net) — intentional?


shivarammysore:
Hi — while linting `src/id_queue.sv` with Verilator (`--lint-only`), I noticed `empty` is assigned at line 193 (`assign empty = &linked_data_free;`) and read at line 410 (`assign empty_o = empty;`), but it is never declared as a `logic`/`wire`. Verilator therefore creates it implicitly and emits an `IMPLICIT` warning.

The signal is a 1-bit reduction, so the implicit net is functionally correct — it just trips strict-lint flows (and `default_nettype none`).

Is relying on the implicit net intentional, or would you welcome a one-line `logic empty;` declaration? Happy to open a small PR. Thanks for common_cells!

phsauter:
This is definitely unintentional.  
More interestingly though the new `v2-dev` branch added a bunch more linting but this hasn't come up so far so we will need to check this specifically in the linting rules as well.  
If you open a PR please do so onto the `v2-dev` branch as we are currently in the final phase of releasing a major (breaking) change that cleans up many other issues surfaced over the years and improves linting and verification across the board.

shivarammysore:
Thanks @phsauter — that matches what we saw. On our side this is non-blocking: we mirror `common_cells` read-only and waive the warning at the build level (`-Wno-IMPLICIT`), keeping the `rtl/` tree a pristine upstream mirror, so there's no urgency for us.

Given `v2-dev` is mid-release and reworks the linting rules anyway, we'll hold off on a PR and instead re-test against `v2-dev` once it's mainstreamed — at which point we can confirm whether the new lint rules already catch (and the cleanup already fixes) the implicit `empty` declaration. Happy to send the one-line `logic empty;` PR onto `v2-dev` if it's still present then.

colluca:
Hi @shivarammysore, @phsauter,

The reason this specific concern does not trigger a CI failure in `v2-dev` is because we are running the Verilator elaboration with the `-Wno-fatal` flag. I find it reasonable to silence _warnings_, as that is the only semantic difference I see w.r.t. _errors_. However, on the specific warning you point out I agree, this can (and should) be easily resolved.

If I remove `-Wno-fatal` there are plenty module elaborations failing, with the following warning classes:
- [`UNSIGNED`](https://verilator.org/guide/latest/warnings.html#cmdoption-arg-UNSIGNED)
[...]

PR #328, opened 2026-07-01
Fix Verilator IMPLICIT, UNOPTFLAT and MULTIDRIVEN warnings

colluca:
Addresses https://github.com/pulp-platform/common_cells/issues/323 as discussed therein.

**TODO**
- [x] Fix [UNOPTFLAT](https://verilator.org/guide/latest/warnings.html#cmdoption-arg-UNOPTFLAT) warnings. @phsauter could you have a look at this since you're more familiar with the CDC modules? My feeling is we might have to waive the warning within those files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,7 +31,6 @@ $(VLT_ELAB_TARGETS): vlt-elab-%: .bender/.checkout
 	$(VERILATOR) --cc \
 		$(shell $(BENDER) script verilator) \
 		--top-module $* \
-		-Wno-UNSIGNED \
 		-Wno-WIDTHEXPAND \
 		-Wno-WIDTHTRUNC \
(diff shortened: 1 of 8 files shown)


phsauter:
Done, I also removed some unneeded waivers and fixed the one UNSIGNED so we can also turn that back on.  
I refuse to fix the width expand and truncate though. It triggers on cases that are 100% clear in their functionality and it is explicitly covered in the SystemVerilog standard how this should work, it should not be a warning.

colluca:
> I refuse to fix the width expand and truncate though. It triggers on cases that are 100% clear in their functionality and it is explicitly covered in the SystemVerilog standard how this should work, it should not be a warning.

Totally agree :grin:

PR #333, opened 2026-07-02
cdc_fifo_gray: split clock domains, configurable sync stages

phsauter:
Forwards SyncStages into the gray FIFO sides and splits pointer/data handling into clearer source and destination path modules. Adds timed gray FIFO test coverage with async delay injection.


colluca (approved):
LGTM :)

PR #334, opened 2026-07-02
cdc_2phase_clearable: rename clock domain modules

phsauter:
Renames the existing split side wrappers to the clearer _src / _dst naming scheme. Also adds section headers in the clearable 2-phase CDC testbench for readability.


colluca (approved):
LGTM :)

PR #335, opened 2026-07-02
cdc_4phase: forward sync stages parameter

phsauter:
Forwards the top-level SyncStages parameter into the 4-phase source and destination halves. Adds timed async-boundary test coverage for the 4-phase CDC.


colluca (approved):
LGTM :)

PR #336, opened 2026-07-02
cdc_2phase: make sync stages configurable

phsauter:
Makes SyncStages configurable for cc_cdc_2phase and forwards it into the source and destination synchronizers. Adds timed async-boundary test coverage for the 2-phase CDC.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -46,12 +46,14 @@ sg-lint:
   timeout: 10min
   script:
     - |
-      if [ "$TOPLEVEL" = "cc_cdc_fifo_gray_tb" ]; then
-        ./test/simulate_cdc_fifo_gray.sh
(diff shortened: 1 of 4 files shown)


.gitlab-ci.yml
```diff
@@ -63,6 +65,7 @@ tests:
       - TOPLEVEL:
           - cc_addr_decode_tb
           - cc_cb_filter_tb
+          - cc_cdc_2phase_tb
           - cc_cdc_2phase_clearable_tb
```
colluca, line 72: I guess alphabetical order is in some sense wrong, unless you interpret `_tb` as a suffix that does not count towards the ordering. But I'd keep the ordering definition naive 😁

colluca (approved):
LGTM apart from following nitpick :)

PR #347, opened 2026-07-16
trip_counter: Fix `CounterExceedsBound` assertion

colluca:
Previously the `CounterExceedsBound` assertion would fire any time the loop would trip. This would happen because `q_o==bound_i` during the last loop iteration, and thus, whatever value `delta_i` assumes in the last loop iteration (apart from zero), `q_o+delta_i` will necessarily exceed `bound_i`.

This PR corrects that assertion, and also documents the expected behaviour of this module more extensively.

### DIFF
--- a/src/cc_trip_counter.sv
+++ b/src/cc_trip_counter.sv
@@ -10,8 +10,38 @@
 //
 // Author: Luca Colagrande <colluca@iis.ee.ethz.ch>
 //
-// Counter which resets automatically when it reaches a specified bound, i.e.
-// when it "trips". Useful e.g. for implementing hardware loop logic.
+// Counter which resets automatically after it reaches a specified bound, i.e.
+// when it "trips". Useful e.g. for implementing hardware loop logic, where the
+// counter can be used to track the induction variable of a loop.
+//
+// Example use case
+// ================
+//
+// Set `delta_i` to `j_incr` and `bound_i` to `j_bound`, and `q_o` will
+// track `j` in the following loop nest:
+//
+//   for (int i = 0; i <= i_bound; i+=i_incr)
+//     for (int j = 0; j <= j_bound; j+=j_incr)
+//
+// Step the `j` loop iterations by asserting `en_i` for each iteration.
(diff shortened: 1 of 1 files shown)


src/cc_trip_counter.sv
```diff
@@ -47,6 +77,9 @@ module cc_trip_counter #(
     assign last_o = (q_o == bound_i);
     assign trip_o = last_o && en_i;
 
-    `ASSERT(CounterExceedsBound, !(en_i && (q_o + delta_i) > bound_i))
+    // Only flag a real overshoot, i.e. incrementing past `bound_i` without landing on it exactly.
+    // On the trip cycle itself (`last_o`), `q_o` already equals `bound_i` and `en_i` is expected to
+    // be high, so `q_o + delta_i` legitimately exceeds `bound_i` there without indicating a bug.
+    `ASSERT(CounterExceedsBound, !(en_i && !last_o && (q_o + delta_i) > bound_i))
```
phsauter: `q_o`, `delta_i` and `bound_i` are all the same width, therefore `(q_o + delta_i)` can overflow back to near-zero which then means it is smaller than `bound_i` (even though it should presumably have tripped).  
So this assertion does not cover everything, the inner operation needs to be extended with zeros (cast to a larger bitwidth) and then summed to avoid this overflow.

phsauter: Oh also it needs an additional guard for the synchronous clear: `en_i && !last_o && !clr_i &&`

phsauter:
```systemverilog
`ASSERT(CounterExceedsBound, !(en_i && !last_o && !clr_i && ({1'b0, q_o} + {1'b0, delta_i}) > {1'b0, bound_i}))
```
This for sure does the trick (probably more than necessary).

phsauter (changes requested):

src/cc_trip_counter.sv
colluca: Good catch!

phsauter (approved):

Issue #288, opened 2026-04-17
Add testbenches or formal verification for each IP


phsauter:
This is a 'best possible' kind of task and not a fixed required goal.  
Ideally all simple IPs should have formal verification, all CDCs should have timed testbenches (maybe even with synthesis to an open technology and netlist simulation) and all other IPs should have a reasonably exhaustive testbench that covers all specified behavior including edge cases.

As much as possible should run with open-source simulators but it is not a requirement.

briann-bui:
Hi @phsauter, I’d like to contribute a focused subtask to this issue by adding formal verification for `cc_heaviside`. It currently appears to have no dedicated testbench or formal harness.

My plan is to add an open-source SymbiYosys/Yosys-compatible formal setup, verify the generated mask across valid inputs and boundary cases, test representative parameter widths including non-power-of-two widths, and integrate it with the existing `formal/` Makefile and documentation.

Would this be a useful scope for #288, or would you prefer that I open a separate child issue for it?

briann-bui:
I’ve opened #350 for the focused `cc_heaviside` formal-verification scope. I do not have permission to attach it as a sub-issue of #288; could a maintainer please add the relationship?

repowazdogz-droid:
Hi @phsauter, I was about to ask about contributing formal verification for `cc_ecc_encode` / `cc_ecc_decode` under #288, then saw #354 reworking the formal setup. I don't want to cut across that, so I'd rather ask first than send anything.

This isn't a speculative idea. I built a scratch prototype locally to check it was practical before proposing it: four cases taken from the truth table in the `cc_ecc_decode` header, covering no error, a single flipped bit inside the Hamming codeword, a flipped extended parity bit, and any two flipped bits. I used negative controls to confirm each case actually fails when the corresponding logic is broken, and the properties proved unbounded with PDR up to `DataWidth=64`. I've deliberately kept it away from any decisions #354 is making about structure, since those are yours to settle.

As far as I can tell these two modules have no functional verification today: `test/cc_ecc_synth.sv` instantiates them with `data_i` unconnected, so it elaborates rather than checks. I couldn't find ECC work in #354 or anywhere else in the tracker.

Would ECC be a useful subtask once #354 lands and the conventions settle, so I can follow whatever shape you end up with? Or is there a module you'd rather see covered first? Happy to wait either way.

phsauter:
[...]

PR #351, opened 2026-07-23
formal: add SymbiYosys proof for cc_heaviside

### LINKED ISSUE #350: Add formal verification for cc_heaviside
Parent issue: #288

## Summary

Add open-source formal verification for `cc_heaviside`, which currently has no dedicated testbench or formal harness.

## Proposed scope

- Add a SymbiYosys/Yosys-compatible formal harness and properties for `cc_heaviside`.
- Prove that `mask_o` contains all and only the asserted bits in the `[0, x_i]` interval for every valid input.
- Exercise boundary inputs and representative parameterizations, including non-power-of-two widths.
- Integrate the new target into the existing `formal/` Makefile and document how to run it.

## Acceptance criteria

- The properties pass with the documented open-source formal flow.
- Boundary cases such as the first and final valid indices are covered.
- Existing formal checks continue to pass.

I would like to work on this focused subtask.

briann-bui:
## Summary

Add open-source formal verification for `cc_heaviside` (Closes #350, subtask of #288).

### Formal verification

- Add `formal/cc_heaviside_properties.sv`, a bound property checker that proves full functional
  equivalence against a loop-based reference model.
- Add `formal/cc_heaviside_formal.sv`, an elaboration harness covering widths 1 through 33,
  including power-of-two and non-power-of-two configurations.
- Add `formal/heaviside.sby`, a depth-1 SymbiYosys proof using the `smtbmc` engine.
- Add `heaviside.check` to `formal/Makefile`.

### Documented behavior

- Document that inputs greater than or equal to `Width` saturate `mask_o` to all ones.
- Capture the same saturation behavior explicitly in the formal reference model.

## How to run

```bash
cd formal
make heaviside.check
```

## Validation

- The property covers widths 1 through 33, including out-of-range encodings for
  non-power-of-two widths.
- Verible lint passes.
- The formal harness compiles and elaborates successfully with VCS X-2025.06.

### DIFF
--- a/formal/Makefile
+++ b/formal/Makefile
@@ -26,7 +26,7 @@ FORMAL_BUILDDIR ?= build
 FORMAL_FLIST := $(FORMAL_BUILDDIR)/cdc_reset_ctrlr.flist
 
 all: fifo.check counter.check fall_through_register.check lzc.check \
-	cdc_reset_ctrlr_half.check cdc_reset_ctrlr_composed.check
+	cdc_reset_ctrlr_half.check cdc_reset_ctrlr_composed.check heaviside.check
 
 $(FORMAL_BUILDDIR):
 	mkdir -p $@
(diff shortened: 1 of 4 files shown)


phsauter (changes requested):
As far as I can tell Property 1 already proves the complete function, Properties 2–6 are logically redundant.  
It can make sense to add redundant checks but really only if one property would be difficult to debug due to complex temporal effects. For this small, combinational proof I do not think they are needed.

Please also clarify the behavior for non-power-of-two `Width` values. Since `idx_width(Width)` can represent values greater than or equal to `Width`, the current unconditional equivalence property effectively makes it so that all out-of-range indices produce an all-ones mask. I actually think this saturation is good and makes it more like a finite-width heaviside function and it preserves monotonicity but it should be documented (in the module as well, we should just make the behavior explicit).

briann-bui:
@phsauter, I addressed your review in commit `337ff0d`:

- removed the redundant Properties 2–6, leaving the full functional equivalence property;
- documented that `x_i >= Width` saturates `mask_o` to all ones in `cc_heaviside`;
- documented the same behavior in the formal checker and reference model;
- updated the PR description to match the revised proof.

Validation: Verible lint passes, and the formal harness compiles/elaborates successfully with VCS X-2025.06. The local SymbiYosys run could not be repeated because this environment does not provide the required `sby`/Yosys-Slang toolchain.

Could you please review again when convenient? Thanks!

phsauter (approved):
Looks good to me, thanks for the help!

phsauter:
> The local SymbiYosys run could not be repeated because this environment does not provide the required `sby`/Yosys-Slang toolchain.

Btw you can get all the open-source tools even with the open-source PDKs in a nice docker container, see here: https://github.com/iic-jku/IIC-OSIC-TOOLS  
This is super useful for local work on laptops etc and also very useful to give AI agents etc.

Issue #356, opened 2026-08-11
cc_stream_fifo_optimal_wrap: Assigning unknown state to `usage_o`


gbellocchi:
The `cc_stream_fifo_optimal_wrap` assigns an unknown state (`X`) to the `usage_o` when `Depth == 32'd2`. This issue was first found when running Spyglass checks into the `snitch_cluster` ([PR #330](https://github.com/pulp-platform/snitch_cluster/pull/330)).

phsauter:
Fixed with #357. I will also start an internal discussion on how we should deal with unknowns.

PR #357, opened 2026-08-12
fix: Drive `usage_o` low if not supported

gbellocchi:
**Problem**
This PR includes a solution for issue #356. The `cc_stream_fifo_optimal_wrap` originally assigned an unknown state to `usage_o` when `Depth == 32'd2`. This issue was first found when running Spyglass checks into the `snitch_cluster` ([PR #330](https://github.com/pulp-platform/snitch_cluster/pull/330)).

**Solution**
This fix simply defines the signal state and drives it low if not supported. The specific value of `usage_o` should not be problematic in this case since the port is generally not connected when instantiated at a higher hierachy.


phsauter:
Thats an easy PR to merge ^^  
Thanks for catching this!

PR #358, opened 2026-08-19
FuseSoC: Sync common_cells.core with Bender.yml 

imchenwu:
1. `cc_stream_delay`: Explicitly zero-extend the LFSR output to the counter width. Connecting the narrow port directly leaves the upper bits undriven (Z) in xsim.
2. Align the filelist in `common_cells.core` with `Bender.yml`.

TODO:

- [ ] [PR in tech_cells_generic](https://github.com/pulp-platform/tech_cells_generic/pull/49) is merged into the newest release.


phsauter:
Odd, as far as I know port connections are an assignment-like context and assignments should always zero-extend in SystemVerilog.  
I will have to check the spec, can you please try if you encounter the same problem with `assign counter_load = lfsr_out;` instead of `assign counter_load = CounterBits'(lfsr_out);`?  
(implicit zero extending instead of explicit extending)

Either I am wrong about the spec of `xsim` is mistakenly not treating port connections as assignment-like.

imchenwu:
Hi Philippe, I also ran the unmodifed `cc_stream_delay` (direct port connection, without intermediate signal) through the full AXI xsim flow on vitis 2025.2 (a newer version compared to the version I previously used) and it works. Therefore I think the original code is spec-compliant and I can revert the logic I introduced in this pr.

Issue #353, opened 2026-08-07
cc_sync_wedge: uses unknown/deprecated `pulp_clock_gating`


fischeti:
Basically, what the title says. I wanted to update the CLINT to common_cells v2 and it currently uses the `sync_wedge` with a 6 year old todo to replace it with common_cells v2 https://github.com/pulp-platform/clint/blob/fc7a1ae5ddbbbc76e5c71c3b184e50162b07d5ac/src/clint.sv#L100 🙃

But the current `cc_sync_wedge` is using `pulp_clock_gating` which I guess has been deprecated and is now an unknown module. should this be replaced with a `tc_clk_gating` instead?

phsauter:
Closed with #359.

PR #359, opened 2026-08-20
sync_wedge: Remove deprecated clock gate

phsauter:
Addresses #353 
The clock gate is not functional since we have a `FFLARNC` register and it is always gated by the enable there.


phsauter:
@fischeti can you quickly think this through as well? I think it is correct since the gated clock only goes to the register and that one already has an enable.

fischeti:
I am a bit confused about the use of the clock gating cell here. Shouldn't the behaviour be that when `en_i` is low, both `f_edge_o ` and `r_edge_o` remain low? Because I think this is currently not given. `en_i` just gates updates to `serial_q` so technically, you could have `f_edge_o ` and `r_edge_o` always high in that case.

fischeti:
sync_wedge

fischeti:
I am not sure but my suspicion is that the previous module was a bit broken. The [one in the CLINT](https://github.com/pulp-platform/clint/blob/fc7a1ae5ddbbbc76e5c71c3b184e50162b07d5ac/src/clint.sv#L100-L142) makes more sense to me, which does not have a clock gating cell and also no enable. Because I am not sure the `sync_wedge` needs an enable at all.

phsauter:
We can change the functionality for v2 and I tend to agree that the current implementation is very odd.

fischeti:
Yes, I agree. I would drop the clock gating and enable signal alltogether, now that we have the opportunity to change the behaviour. The clear signal might be a bit tricky, because we cannot clear the `tc_sync` and if we only clear `serial_q`, you might cause a pulse in the cycle after you clear, which is probably undesired.

phsauter:
Re clear signal, we already made the decision to exclude it from clock domain crossings and similar things so we can easily also remove it here.  
I also had Codex check all other PULP projects as well as OpenHWGroup and all of them tie the enable, so I think its fine to remove it.

fischeti (approved):
LGTM, Thanks!

Issue #360, opened 2026-09-04
Backport #328 (Fix Verilator IMPLICIT, UNOPTFLAT and MULTIDRIVEN warnings) to v1.40


luposmi:
In #328, several Verilator warnings that can cause builds to fail when warnings are treated as fatal were fixed.

Users of the 1.x branch currently need to either apply these fixes themselves or disable the corresponding fatal warnings. If v1.40 is intended to be the final release before v2.0, I think it would be useful to backport the relevant fixes to the 1.40 branch and potentially make a v1.40.1 release.

Additionally, common_cells.core in v1.40.0 still declares the package version as 1.39.0.

Edit:
If interested, my cherry pick looks like this: https://github.com/luposmi/common_cells/tree/backport-328.
Feel free to use this as a baseline, but note that it does not adress the version mismatch in the core file.

I also noticed that #328 removed a lint_off which was not fixed by the commit (which therefore lead to errors with certain parameters). The given branch fixes this and I created a pull request to reintroduce the change to master in #361 .

phsauter:
Good notes and yes the intention was for 1.40 to be the last one. I think we need a 1.40.1 to fix this issues though.

PR #361, opened 2026-09-04
rr_arb_tree: Reintroduce lint_off for SPLITVAR

luposmi:
I noticed this issue while working on #360. There, i already fixed this. 

They where removed in e29d251. However, they did not address the underlying issue (if NumIn:=2, trhen NumLevels = $clog2(2) = 1.Then, the arrays cannot be split as they only contain one element). I assume they where removed by accident.


phsauter:
Yes I am pretty sure you are correct and the explanation makes sense.

PR #362, opened 2026-09-15
Add `bender` target to suppress deprecation warnings

da-gazzi:
Migrating large designs with many layers of nested dependencies to `common_cells` is not trivial. To avoid CI logs becoming completely unusable because every `fifo_v3`, `FFLARN` etc emits its own deprecation notice, I submit this PR, which adds a target (`cc_no_deprecated_warnings`) to suppress the deprecation warnings.


phsauter:
@colluca I think this is a good idea but I am a bit worried that people will just slap it in everywhere and with that completely ignore that they are using deprecated things. How do you see it?

Bender.yml
```diff
@@ -169,6 +169,10 @@ sources:
   # Included by default for backward compatibility. Set the target 'cc_no_deprecated'
   # to exclude them and enforce use of the new cc_* names.
   - target: not(cc_no_deprecated)
+    defines:
+      # Suppress migration warnings without removing the compatibility wrappers.
+      COMMON_CELLS_NO_DEPRECATED_WARNINGS:
+        target: cc_no_deprecated_warnings
```
colluca, line 175: I don't understand this syntax. Does bender now allow that a define is filtered both by an outer target key and an inner target key?

Ultimately, what I want to understand is what is the default behaviour.

colluca (commented):
I think this is reasonable, so long as it is not the default behaviour; this should allow use of deprecated cells, but reporting it with a warning.

In the end, I believe it to be in a user's best interest to transition to the non-deprecated cells, and I thus wouldn't make it our duty to police it. But that's just my subjective preference.

Bender.yml
da-gazzi: @colluca Yes, bender can filter defines by target. This is ultimately a workaround for the lack of a global per-target define. See https://pulp-platform.github.io/bender/sources.html#target-conditional-include-dirs-and-defines

Default behavior is that the warnings are emitted.

phsauter:
Okay I tend to agree, then we can merge it.

PR #363, opened 2026-09-17
hw: Add new common cells to down/up size data stream

Lore0599:
# Add cc_stream_downsizer / cc_stream_upsizer
This PR introduces two generic ready/valid stream width converters, parametrized.

## New Cells 
- **cc_stream_downsizer** drains one wide beat into `WideWidth/NarrowWidth` narrow beats, LSB first. It doesn't register the wide word: as long as `inp_ready_o` is low, the master must hold it stable anyway (standard handshake rule), so each slice is just indexed straight off `inp_data_i`.

- **cc_stream_upsizer** does the reverse. Here each narrow beat is genuinely new data, so the first `Ratio-1` slices do need to be captured into a register bank. The last slice is the exception: it's streamed out
combinationally together with the stored ones; in the same cycle it's accepted, instead of being registered and presented a cycle later. That keeps its acceptance directly tied to `oup_ready_i`, and means the sequence "accept last slice of one word" / "start accepting the first slice of the next" costs zero extra cycles.

## HW cost: 
Downsizer is register-free besides the counter; upsizer holds `(Ratio-1) * NarrowWidth` bits of state. Neither adds combinational depth beyond `cc_counter`'s own.

## Testbench
Add two testbenches generated by AI 🤖


src/cc_stream_downsizer.sv
```diff
@@ -0,0 +1,60 @@
...
+  ) i_slice_cnt (
+    .clk_i,
+    .rst_ni,
+    .clr_i     (inp_ready_o),
+    .en_i      (slice_en),
+    .load_i    (1'b0),
+    .down_i    (1'b0),
+    .d_i       ('0),
+    .q_o       (slice_q),
+    .overflow_o(  /* unused */)
```
colluca: nitpick: spacing between parenthesis and comment

src/cc_stream_downsizer.sv
colluca, line 1: I think a trip_counter would be even better, it basically wraps the `last` and `clear` logic inside the module for you. Same also for the upsizer.

src/cc_stream_upsizer.sv
```diff
@@ -0,0 +1,78 @@
...
+    );
+
+    always_ff @(posedge clk_i or negedge rst_ni) begin
+      if (!rst_ni) begin
+        data_q <= '0;
+      end else if (accept && !last_slice) begin
+        data_q[slice_q*NarrowWidth+:NarrowWidth] <= inp_data_i;
+      end
+    end
+  end
```
colluca: Use the assertion macros, e.g. `` `FF``, and an explicit `data_d` signal which is default-assigned to `data_q` in an always_comb block. Probably overcautious, but it's good practice to make sure latches aren't introduced.

test/cc_stream_downsizer_tb.sv
```diff
@@ -0,0 +1,142 @@
...
+    forever begin
+      word = {$urandom(), $urandom()};
+      repeat ($urandom_range(0, 4)) @(cb);
+      cb.inp_data  <= word;
+      cb.inp_valid <= 1'b1;
+      @(cb);
+      while (!cb.inp_ready) @(cb);
+      cb.inp_valid <= 1'b0;
+    end
+  end
```
colluca: There's quite some generic logic here, that's repeated also in the upsizer testbench.
And much of it already exists in common_verification which is already a dependency of common_cells, e.g. https://github.com/pulp-platform/common_verification/blob/master/src/rand_stream_mst.sv.

I would reuse at least the clock/reset driver and the randomized stream driver.

colluca (changes requested):

src/cc_stream_downsizer.sv
Lore0599: done

src/cc_stream_downsizer.sv
Lore0599: done

src/cc_stream_upsizer.sv
Lore0599: yes makes sense, done!

src/cc_stream_downsizer.sv
```diff
@@ -0,0 +1,58 @@
...
+    .Width(CntWidth)
+  ) i_slice_cnt (
+    .clk_i,
+    .rst_ni,
+    .clr_i  (1'b0),
+    .en_i   (slice_en),
+    .delta_i(CntWidth'(1)),
+    .bound_i(CntWidth'(Ratio - 1)),
+    .q_o    (slice_q),
+    .last_o (  /* unused */),
```
colluca: Spurious spaces are still there :grin:

src/cc_stream_downsizer.sv
```diff
@@ -0,0 +1,58 @@
...
+  input  logic                   inp_valid_i,
+  output logic                   inp_ready_o,
+  /// Narrow output stream
+  output logic [NarrowWidth-1:0] oup_data_o,
+  output logic                   oup_valid_o,
+  input  logic                   oup_ready_i
+);
+
+  logic [CntWidth-1:0] slice_q;
+  logic slice_en;
```
colluca: nitpick: no need for this intermediate signal (used only once, and not really documenting what it does), just inline `inp_valid_i && oup_ready_i`

src/cc_stream_upsizer.sv
```diff
@@ -0,0 +1,77 @@
...
+  /// Data width of the narrow (input) side, in bits.
+  parameter  int unsigned NarrowWidth = 32'd1,
+  /// Data width of the wide (output) side, in bits. Must be an integer multiple of `NarrowWidth`.
+  parameter  int unsigned WideWidth   = 32'd1,
+  /// Number of narrow beats per wide beat (derived, do not override).
+  localparam int unsigned Ratio       = WideWidth / NarrowWidth,
+  /// Width of the internal slice counter (derived, do not override).
+  localparam int unsigned CntWidth    = cc_pkg::idx_width(Ratio),
+  /// Width of the register bank holding the non-final slices (derived, do not override).
+  localparam int unsigned StoreWidth  = (Ratio - 1) * NarrowWidth
```
colluca: Not used in port list. Prefer localparam definitions in module body. Same in downsizer.

colluca (changes requested):
Few minor comments left

colluca (approved):

phsauter (approved):
Assertions are working and testbenches as well, everything looks good.

